PR #3, opened 2020-02-04
Minor cleanup, mostly comment-level

jaytmiller:
Cleaned up processing code:
   Renamed suffixes attribute to download_suffixes.
   Replaced unassoc_files method copy with true method; head off potential bugs.
Cleaned up planner:
   Documented core "plan" function for planning one ipppssoot.


jaytmiller:
Just comments,  I tested anyway with a job.  I won't self-merge for anything substantive but this is a waste of your time.

PR #27, opened 2020-10-14
added terraform deployment scripts

bhayden53:
Ran 10 datasets on infrastructure made by these scripts, they failed in caldp but the infrastructure seemed to work fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,9 @@ build/
 sdist/
 dist/
 
+terraform/.terraform
+terraform/*.tfvars
+terraform/*.tfstate*
+terraform/*.txt
+terraform/*.out
+
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,8 @@ sdist/
 dist/
 
 terraform/.terraform
-terraform/*.tfvars
 terraform/*.tfstate*
 terraform/*.txt
 terraform/*.out
+terraform/local.tfvars
 
--- a/terraform/Network_IAM.tf
+++ b/terraform/Network_IAM.tf
@@ -1,11 +1,3 @@
-variable "batchvpc_cidr" {
-  type = string
-}
-
-variable "batchsn_cidr" {
-  type = string
-}
-
 resource "aws_vpc" "batchvpc" {
   cidr_block = var.batchvpc_cidr
   enable_dns_hostnames = true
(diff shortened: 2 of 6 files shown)


TheRealZoidberg (approved):

TheRealZoidberg (approved):

TheRealZoidberg (approved):
A whole lot of code! But I do not see anything obvious.

bhayden53:
> A whole lot of code! But I do not see anything obvious.

Most of it is IAM and networking....

jaytmiller (commented):
This looks great to me overall but I had a few comments/suggestions:

1. I think there's a coding convention of creating a variables.tf file where you put every variable you need to configure a module.   This gives deployers one place to look.   There are similar conventions for creating a "local.tf" and "outputs.tf" module where you put locals and outputs respectively,  but add only if you have  some.

2. Instead of documenting "fill out x + y + z" create  a template.tfvars file or something and define it as something close to what you got working;  close == don't disclose sensitive information,  e.g. account number.

3. Personally I wouldn't use "local.tfvars" because I believe there is a convention of defining local.tf which contains "local" definitions (global vars defined by expressions).   I could be wrong.    Look down inside .terraform for example code.

4. ITSD will probably push back on role and VPC creation,  they did for JupyterHub.   So that may need to be  restructured eventually to (a) import their network resources using "data" and (b) control role creation with permission boundaries where applicable. 

5. If/when you do start importing network resources,  you can make variables which define names  or name  patterns  and then filter  resources on those,  avoiding the need  to meticulously cut-and-paste AWS ids.   Conceivably these patterns 
 can be invariable between accounts as well but in any case the patterns should vary regularly.   spacetelescope/terraform-deploy has some examples of that,  one pattern picks up 3 DMZ subnets,  one 3 public subets,  one the VPC.   It comes in handy if/when  IDs start multiplying when you e.g. distribute batch across multiple AZs/subnets ITSD defines.

6. Maybe make a variable for "batch_deployment" so that multiple instances can be instantiated in one account.  Feed it into all the resource names.

7. user_data is a trip!   Potentially we can create something maintainable by combining this:  https://www.tommyblue.it/2020/02/22/how-to-use-a-launch-template-in-aws-batch-compute-environment-with-custom-user_data-file/  and  this: https://www.bogotobogo.com/DevOps/Terraform/Terraform-terraform-userdata.php.

bhayden53:
I tried to separate out the parts that I thought may eventually either directly get made by ITSD, or they would at least need to review for permissions boundaries, into the one file. I was anticipating that they would eventually take over, but I also thought this would be a useful way to tell them what we need.

I will make some of the restructuring changes today, thanks for highlighting the proper convention.

bhayden53:
I made a ticket for refactoring the user_data situation, thanks for looking into that and providing links.

PR #28, opened 2020-10-15
refactored calcloud commands for current caldp calling sequence

bhayden53:
Changed the command sequence to make calcloud compatible with current caldp calling sequence. Confirmed that calibration jobs are succeeding on AWS Batch with this current version and caldp master branch.


jaytmiller (approved):
Nice,  LGTM

PR #32, opened 2020-12-02
Terraform remote

alphasentaurii:
Added code block for terraform s3 remote backend configuration. Removed "profile" from provider block. Modified Region to be a variable for setting in .tfvars files instead of hardcoded.


bhayden53:
Can you add .DS_Store to .gitignore and remove it from the commit?

PR #33, opened 2020-12-11
Terraform/dev

bhayden53:
updated terraform for the situation where the network and IAM is already set up by the security team. The roles and arns are now variables that are passed to the batch objects, rather than being created and managed by terraform. 

This does make it a little harder for someone to pick this up and stand up a full environment from nothing; we should consider putting the Network_IAM.tf objects back into the repo in such a way that users could re-institute that model.


bhayden53:
going to let @jaytmiller  PR first, because there will be some merge issue between our PR's, and it's probably easier for me to work off of his changes than the other way around.

PR #36, opened 2021-01-28
Unique  resource naming + SSM parameters

jaytmiller:
Draft Terraform changes to make resource names unique enabling multiple conflict-free deployments per account.

Configuration for all resources created by ITSD is now input via SSM
parameters leaving only region, environment, and image_tag variables, all of
which have reasonable default values.  environment is also defined by
default for each account, so effectively a default calcloud build can
be completely parameter free using only the assumptions region=us-east-1
and image_tag=latest.

Among the resources now defined by SSM parameters are the Batch AMI ID,
VPC, and subnets.

This update adds three new files:

1. parameters.tf which is used  to define SSM parameters much like
varaibles.tf defines variables.  (Made up convention)

2. locals.tf which defines a few computed (local) values in one place
so they can be referenced simply everywhere else.  (Standard convention)
As an example,  these enable the SSM value of environment to be
overridden using a variable,  and also tweak "-" handling so the ""
environment is not seen as an orphan trailing slash.

3. outputs.tf which defines values  which should be printed out after
each Terraform run. (Standard convention)

### DIFF
--- a/terraform/remote_state.tf
+++ b/terraform/remote_state.tf
@@ -1,0 +1,3 @@
+terraform {
+  backend "s3" {}
+}
--- a/terraform/remote_state.tf
+++ b/terraform/remote_state.tf
@@ -1,3 +1,0 @@
-terraform {
-  backend "s3" {}
-}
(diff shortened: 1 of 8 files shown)


terraform/remote_state.tf
```diff
@@ -1,3 +0,0 @@
-terraform {
```
bhayden53, line 1: Did you mean to undo this before PR? I don't see that it was moved anywhere else.

PR #37, opened 2021-01-29
lambda s3 event triggers for job submission

bhayden53:
- lambda function triggers on s3 put for placed- messages
- lambda uses env vars set automatically from terraform to set inputs/outputs/queues/job-definitions
- calcloud requires several env variables for job submission (since job submission is run from the lambda environment)
- minor terraform tweaks to improve reliability when redeploying would replace existing compute environment
- changed hardcoded instance types to "optimal" to allow Batch to optimize by price/cpu/memory, and changed allocation strategy to "BEST_FIT_PROGRESSIVE"


calcloud/lambda_submit.py
```diff
@@ -0,0 +1,37 @@
...
+    input_path = f"{bucket}/inputs"
+    planned_resource_tuples = plan.get_resources_tuples(ipppssoots, bucket, input_path)
+
+    # reproduces the printed output of provision
+    provisioned_resource_tuples = provision.get_plan_tuples(planned_resource_tuples)
+    
+    s3_client = boto3.resource('s3')
+
+    # submits the jobs
+    for p in provisioned_resource_tuples:
```
jaytmiller, line 26: Just as a reminder,  this loop will probably need to be limited to a few thousand ipppssoots.  

It might be good to reject lengthy requests or maybe make it recurse to sub-lambdas?

jaytmiller (approved):
Other than possible ipppssoot limits and/or making the lambda recurse in blocks of 1k-5k ipppssoots somewhere,  LGTM

calcloud/lambda_submit.py
bhayden53: Because of the S3 event trigger this loop will actually only ever run on a single ipppssoot now, because the events trigger on each placed-ipppssoot message individually.

PR #40, opened 2021-02-16
Calcloud 194 rescue

jaytmiller:
This doesn't have memory handling yet but I was able to rescue one terminated  job and have it successfully process on my string.    I think messaging is now in a fairly usable, tested state and ready for review and/or usage.   Both the lambda and working doctests demonstrate how the messaging and I/O abstractions work.   I think memory handling should be done in a follow-on PR.


calcloud/s3.py
```diff
@@ -216,8 +224,8 @@ def upload_directory(dirpath, s3_dirpath):
     raise NotImplementedError("upload_directory hasn't been implemented yet.")
 
 
-def list_directory(s3_prefix, client=None, max_objects=1, exclude_prefix=True):
-    """Given `s3_dirpath_prefix` s3 bucket and prefix to list, return the full
+def list_objects(s3_prefix, client=None, max_objects=10 ** 6):
+    """Given `s3_dirpath_prefix` s3 bucket and prefix to list, yield the full
```
bhayden53, line 228: a million is a lot, but it's worth mentioning that this could be the source of major headaches if we ever actually have more than a million things to list, as technically it "yields the full s3 paths of the first max_objects objects" in the bucket with the prefix, right?

lambda/JobRescue/rescue_handler.py
```diff
@@ -0,0 +1,51 @@
...
+        bucket_name = event["Records"][0]["s3"]["bucket"]["name"]
+        message = event["Records"][0]["s3"]["object"]["key"]
+        ipst = message.split("-")[-1]
+        # rescue_reason = f"operator posted {message} message"
+        print(f"received {message} on bucket s3://{bucket_name}")
+        assert hst.IPPPSSOOT_RE.match(ipst) or ipst == "all",  "Bad ipppssoot value: " + repr(ipst)
+    except Exception as exc:
+        return dict(statusCode=200, body="Rescue inputs parsing failed with exception: '{str(exc)}'")
+
+    try:
```
bhayden53, line 31: I have a general aversion to putting this much code in a try/except, although I'm not sure that's based on sound principles or not.

In general, proper AWS API calls should send a response code rather than raising an exception, so this should only catch major errors in our wrappers. So I think it should be ok.

terraform/lambda_job_rescue.tf
```diff
@@ -0,0 +1,64 @@
...
+  # for now this role was created by hand in the console, it is not terraform managed
+  lambda_role = data.aws_ssm_parameter.lambda_delete_role.value
+  
+  # lambda_role = data.aws_ssm_parameter.lambda_rescue_role.value
+
+  environment_variables = {
+    # comma delimited list of job queues, because batch can only list jobs per queue
+    JOBQUEUES="${aws_batch_job_queue.batch_queue.name},${aws_batch_job_queue.batch_outlier_queue.name}"
+    S3_PROCESSING_BUCKET=aws_s3_bucket.calcloud.id
+    FILESHARE=data.aws_ssm_parameter.file_share_arn.value
```
bhayden53, line 48: I don't think you need to pass the FILESHARE variable in. That should only be needed if you plan to directly call the refresh cache operation.

In fact I don't think you use any of the env variables. The only problem would be if I made a hard dependency on one of them in any of the calcloud imports. That might be the case, though I don't _think_ it is here.

calcloud/io.py
```diff
@@ -0,0 +1,294 @@
...
+    def __init__(self, s3_path, client=None):
+        """API operating relative to `s3_path` using S3 `client`."""
+        self.client = client or s3.DEFAULT_S3_CLIENT
+        self.s3_path = s3_path
+
+    def path(self, prefix):
+        """"""
+        return self.s3_path + "/" + prefix
+
+    def expand_prefix(self, prefix):
```
bhayden53, line 45: could you add comment docstrings to expand_prefix, expand_all, list_s3, and list?

jaytmiller: Will do.

calcloud/s3.py
jaytmiller: Yeah,  the intent is to yield only the first 10**6 as a safety hedge.   

We could make max_objects configurable in S3Api,  make it  an exception if there are unlisted objects,  then test it with a low threshold.   Or,  just bump it to 10**7.  

Albeit 10**7 = 10**4 *  10**3 => 10**4 * $0.005 => $50.    But that's *if* 10**7 objects actually exist.

I also noticed that GET and SELECT(?) are 8x cheaper than PUT, COPY, POST, DEL.

SELECT seems to treat S3 objects as database tables.   I'm not sure how you handle concurrent  updates.

bhayden53: seems to follow the pattern in other services as well of a 5-10x cost increase for write operations over reads.

lambda/JobRescue/rescue_handler.py
jaytmiller: This was a great comment,  for one I should have used code 400 not 200.   

Secondly,  I'm not sure how much AWS does automatically.  But I'm guessing they don't (1) identify it as "rescue" (2) report ipst.  (3) distinguish input errors vs.  processing errors.  (4) I'm not sure we want a full exception traceback if they do that.      So I'm leaving this as-is after correcting to code 400,  but an alternative might be to just rip most of it out,   let me know if you'd prefer the latter.

terraform/lambda_job_rescue.tf
jaytmiller: OK,  I'll nix FILESHARE.   I thought I'd definitely need S3_PROCESSING_BUCKET,  is that handled some other way?   If I do need it,  should we rename env instances of BUCKET to S3_PROCESSING_BUCKET everywhere else?

lambda/JobRescue/rescue_handler.py
bhayden53: I'm fine with it as-is unless we see problems

terraform/lambda_job_rescue.tf
bhayden53: Yeah, you actually pull in the bucket through the event via
```bucket_name = event["Records"][0]["s3"]["bucket"]["name"]```

I think there's probably still a rogue env variable in one of my lambdas as well for that one.

bhayden53:
LGTM

PR #42, opened 2021-02-18
removed deprecated direct setting of memory and vcpu 

bhayden53:
and used resourceRequirement which could facilitate fargate in the future

good news bad news
good news: could use fargate in future, no longer using deprecated params according to aws docs

bad news: the overrides don't show up anywhere in the console, I had to make sure they were working by hardcoding a small number and watching a job fail


jaytmiller (approved):
LGTM.

PR #43, opened 2021-02-18
cloudWatch event rule that captures failed jobs 

bhayden53:
and sends them to lambda for handling

This includes jobs that were cancelled via the cancel lambda + terminate_job api


jaytmiller (approved):
Looks great to me,  thanks for lining this up and pushing it forward.

PR #44, opened 2021-02-19
Memory fail messaging

jaytmiller:
Work in progress, incomplete.

Start on memory fail  messaging and the control file,  cleanup of the planner and provisioner interfaces simplifying to ipppssoot-by-ipppssoot and clarifying top-level interfaces to preserve if possible.  JobID and retry tracking,   no CloudWatch event or job status query yet.

### DIFF
--- a/calcloud/plan.py
+++ b/calcloud/plan.py
@@ -5,8 +5,6 @@ The idea behind creating resourcess is to generate enough information such that 
 set of ipppssoots can be assigned to well tuned processing resources.
 """
 import sys
-import datetime
-import math
 import os
 from collections import namedtuple
 
@@ -15,15 +13,6 @@ from . import metrics
 from . import log
 
 # ----------------------------------------------------------------------
-
-# def get_batch_name(name):
-#     when = datetime.datetime.now().isoformat()
-#     when = when.replace(":","-")
-#     when = when.replace(".","-")
-#     return name + "-" + when[:-7]   # drop subseconds
-
-# One key function of JOB_DEFINITION appears to be indelible declaration of Docker image.
-# New image == new definition required.
 
 """
 WFC3: 95 per duration 0.93 hr
@@ -42,6 +31,8 @@ STIS: 95 per duration 0.12 hr
 STIS: 99 per duration 0.24 hr
 STIS: 95 per mem 0.64 gb
 STIS: 99 per mem 1.0 gb
+
+
 """
 
 # For optimization with AWS Batch, key insights are:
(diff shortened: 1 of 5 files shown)


jaytmiller:
I completed a first go at this,  grabbing the Batch fail event and implementing automatic rescues for the first 4 tries with status all tracked in the ipppssoot control file.

The submit  lambda passes a "memory_retries" count  into the planner.   My thinking was the planner would implement some memory increase function and continue reporting human readable memory requirements.   Then the provisioner would be tasked with mapping those requirements onto whatever queues, job definitions, etc are needed.     The  planner and provisioner still run on the command line so it should be possible to plug in Ru's model and then see what kinds of requirements  and plans it comes up with...  and potentially write unit tests for that.

calcloud/plan.py
```diff
@@ -65,46 +56,67 @@
...
     """
-    info = list(JOB_INFO[instr.lower()])
+    info = [0] * 3
     try:
         memory_megabytes, cpus, wallclock_seconds = metrics.get_resources(ipppssoot)
         info[0] = cpus
-        info[1] = memory_megabytes + 512    # add some overhead for AWS Batch (>= 32M) and measurement error
-        info[2] = max(int(wallclock_seconds*cpus*6),120)  # kill time,  so too high is better than too low
+        info[1] = memory_megabytes + 512  # add some overhead for AWS Batch (>= 32M) and measurement error
+        info[2] = max(int(wallclock_seconds * cpus * 6), 120)  # kill time,  BEWARE:  THIS LOOKS WRONG
```
bhayden53, line 100: What's wrong here? I think I wrote the current form

jaytmiller: If nothing else,   I don't understand all that fudging myself and I'm responsible for some/all of it.   But I think maybe that should be "/ cpus" to determine time,  not "* cpus".  And I have no clue where "* 6" came from but it may just be a generous fudge factor since it's only a kill time.

bhayden53: So I can tell you what happened on part of the wall clock anyway. Batch wasn't accepting jobs less than 60 seconds, it was erroring on submit, so Ru set the minimum to 60 seconds. Then I found some jobs were exceeding the limit when set at 60 seconds, so I set the minimum limit to 120 seconds.

I actually don't recall off the top of my head what the seconds*cpus*6 means. But it does seem weird, wouldn't more CPUs REDUCE the needed time?

jaytmiller: About CPUs,  exactly.

jaytmiller: Although,  strictly speaking,  straight up "/ cpus" is too optimistic and I wouldn't want to guess what it really is.  It should probably be dropped from the kill time altogether,  if more cpus are faster,  great,  if not,  no harm done.

jaytmiller:
Replaced by the rescue-place-overhaul PR

PR #46, opened 2021-02-20
Add tox code checks

jaytmiller:
Added tox setup for black, flake8, bandit.
Skips pytest and codecov for now.  
No GitHub actions yet.


bhayden53:
LGTM unless you're planning to add more?

PR #47, opened 2021-02-22
initial skeleton of a deployment script

bhayden53:
successfully deployed environment from scratch in sandbox using this deployment script.

### DIFF
--- a/terraform/deploy.sh
+++ b/terraform/deploy.sh
@@ -61,7 +61,6 @@ docker push ${CALDP_DOCKER_IMAGE}
 
 # deploy rest of terraform
 cd ../calcloud-${CALCLOUD_VER}/terraform
-awsudo $ADMIN_ARN terraform plan
 # manual confirmation required
 awsudo $ADMIN_ARN terraform apply
 
--- a/terraform/deploy.sh
+++ b/terraform/deploy.sh
@@ -1,0 +1,69 @@
+#! /bin/bash -xu
+
+# ADMIN_ARN is set in the ci node env and should not be included in this deploy script
+
+# variables that will likely be changed frequently
+CALCLOUD_VER="0.1.0"
+CALDP_VER="0.1.1"
+CAL_BASE_IMAGE="stsci/hst-pipeline:CALDP_20201208_DRZ_final"
+# this is the tag that the image will have in AWS ECR
+CALDP_IMAGE_TAG="latest"
+
+# variables that will be changed less-frequently
+TMP_INSTALL_DIR="/tmp/calcloud_install"
+
+mkdir $TMP_INSTALL_DIR
+cd $TMP_INSTALL_DIR
+
+# calcloud source download/unpack
+wget "https://github.com/spacetelescope/calcloud/archive/v$CALCLOUD_VER.tar.gz"
+tar -xvzf "v$CALCLOUD_VER.tar.gz"
+rm "v$CALCLOUD_VER.tar.gz"
+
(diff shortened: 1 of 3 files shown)


terraform/deploy.sh
```diff
@@ -0,0 +1,69 @@
...
+# this is the tag that the image will have in AWS ECR
+CALDP_IMAGE_TAG="latest"
+
+# variables that will be changed less-frequently
+TMP_INSTALL_DIR="/tmp/calcloud_install"
+
+mkdir $TMP_INSTALL_DIR
+cd $TMP_INSTALL_DIR
+
+# calcloud source download/unpack
```
jaytmiller, line 18: Maybe comment this out.  Given the git clone for the pip install,  there's no point is there?

terraform/deploy.sh
```diff
@@ -0,0 +1,69 @@
...
+# cd ~/bhayden/caldp
+CALDP_DOCKER_IMAGE="${repo_url}:${CALDP_IMAGE_TAG}"
+docker build -f Dockerfile -t ${CALDP_DOCKER_IMAGE} --build-arg CAL_BASE_IMAGE=${CAL_BASE_IMAGE}  .
+# need to "log in" to ecr to push the image
+awsudo $ADMIN_ARN aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin $repo_url
+docker push ${CALDP_DOCKER_IMAGE}
+
+# deploy rest of terraform
+cd ../calcloud-${CALCLOUD_VER}/terraform
+awsudo $ADMIN_ARN terraform plan
```
jaytmiller, line 64: I don't think there's a need to both plan and apply unless you first capture the plan for review,  then use it during the apply.  Otherwise,  the apply plans and presents it for review.  I think this just gives two plans in the output.

jaytmiller (approved):
Couple comments,  but LGTM

terraform/deploy.sh
bhayden53: I like having a completely clean install area each time. It is deleted at the end of the script as well.

terraform/deploy.sh
bhayden53: good call, it's something I do out of habit and I forgot that it's not really required.

terraform/deploy.sh
bhayden53: depending on which exact line you were referring to, I do pull the calcloud release instead of cloning the repo... it's only caldp that needs the pip install. calcloud is needed for terraform.

jaytmiller: Yep,  that  was it,  I didn't notice that there were  two repos installed.

jaytmiller (commented):
LGTM again,  I was just commenting,  you already merged.  :-)

jaytmiller:
Forgot to mention:  this solves the ECR/CALDP chicken and  egg,   right?   Great,  I'm psyched to use this baby!

bhayden53:
> Forgot to mention: this solves the ECR/CALDP chicken and egg, right? Great, I'm psyched to use this baby!

Yes although it's not very "smart" in that it will build/push the caldp image every time it's run. The terraform stuff is idempotent although there is a slight annoyance where terraform replaces the lambda envs every time you apply, due to a timestamp that changes, even if the source code hasn't changed. It makes it a little annoying to look for "real" changes.

The main thing I wanted out of this was an easy way to pull in specific versions of caldp, calcloud, and the base image, reliably.

PR #48, opened 2021-02-23
Rescue place overhaul

jaytmiller:
This is the full PR for rescue,  including automatic rescue/retries for memory failures reported by CloudWatch Batch events.
I refactored the S3 trigger lambda to use both the io module and revised lambda_submit.main.
Maybe the best way to review it is to take the I/O module on faith for now and look at the top level logic exposed in the S3 trigger lambda, rescue lambda,  batch event lambda,  and lambda_submit.main which is now called for both place and  rescue.
There are some black and flake8 change here.
I consolidated calcloud.plan and  .provision into  calcloud.plan and made a single clean entry point called during submissions which now includes the retry counter.   The only important  thing left really is the plan.get_plan(..., retries)  --> Plan() interface.


bhayden53:
I read all the diffs and didn't find anything to complain about, but the changes are I think definitely more that I can easily parse just by reading.

I'm inclined to merge, deploy to sandbox, and use my test data to verify all of the lambdas are working as intended by monitoring messages and logs.

jaytmiller:
As we slacked,   this is ready to proceed  to testing.

PR #50, opened 2021-02-24
Dropped rescue for terminate.

jaytmiller:
One line change,  seemed to work fine killing a submitted job which was RUNNABLE.


bhayden53:
lgtm

bhayden53:
so I was wrong about this. We want to be able to rescue terminated jobs if an operator manually posts a rescue message; we just don't want terminated jobs to auto-rescue for any reason.

PR #53, opened 2021-02-25
fixed bandit error for scrape_batch tmp file

bhayden53:
other bandit errors should be fixed in a future PR by @jaytmiller


jaytmiller (commented):
LGTM

PR #55, opened 2021-03-01
Delete with control file,  rescue terminated no retries,  broadcasting

jaytmiller:
This is a significant overhaul of the delete, cancel, and place lambdas,  batch event handler,  and core lambda_submit  logic.   It also includes library updates to make it easier,  and adds a new broadcast lambda for sending arbitrary lists of messages using a divide-and-conquer approach.

Although the basic premise of using the control file for the JobDelete lambda panned out in greatly simplified code,  it did not appear to be robust enough as an emergency tool because duplicate jobs overwrite the control file losing one job_id.   I added code to delete-all to first list all killable job ids,  then  broadcast a cancel-job_id message for each one.   The cancel-job_id does nothing but terminate the specified orphan,  it does not interact with the control file or messaging which is handled by a parallel/competing cancel-ipppssoot message.  The cancel-ipppssoot message  also terminates  jobs but exceptions from terminate_job are caught.


bhayden53:
LGTM

PR #57, opened 2021-03-05
new HST keywords for cloud software versions

bhayden53:
3 HST keywords are passed through the deploy script into the job definition as environment variables, where they're used by caldp to set the keywords in the raw headers, which pass through to the calibrated products (verified).

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -14,3 +14,21 @@ variable region {
   type = string
   default = "us-east-1"
 }
+
+variable awsdpver {
+  description = "HST keyword for caldp repo tag version"
+  type = string
+  default = "undefined"
+}
+
+variable awsysver {
+  description = "HST keyword for calcloud repo tag version"
+  type = string
+  default = "undefined"
+}
+
+variable csys_ver {
+  description = "HST keyword for docker base calibration image tag"
+  type = string
+  default = "undefined"
+}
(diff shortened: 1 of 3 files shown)


terraform/variables.tf
```diff
@@ -14,3 +14,21 @@ variable region {
   type = string
   default = "us-east-1"
 }
+
+variable awsdpver {
+  description = "HST keyword for caldp repo tag version"
+  type = string
+  default = "undefined"
```
jaytmiller, line 21: This may be your intent,  but if you want  to force the definition of those vars during deployment you can just leave off the default.

jaytmiller (approved):
LGTM

terraform/variables.tf
bhayden53: I actually don't want to! To make it simpler for developers who just want to get something up quick in sandbox. I like the idea of a terraform apply not having any required variables, and I'm going to try to keep it that way as long as I can.

lisaesherbert: HSTSDP-1380 will be used to update the keyword database so that these new keywords are written to at least the raw headers, just like CSYS_VER.  They will have short descriptions at this time.  Brian assures me that caldp [sets] the keywords with fits.setval which is basically an “update or insert” operation. It changes the keyword if it’s already there, and makes a new one if it’s not [and] it does leave the comment as well

PR #58, opened 2021-03-05
Add jobdef ladder

jaytmiller:


jaytmiller:
This is what I have so far for the job def ladder.   

I'm still doing an outlier test to see how it progresses through retries and higher order job defs.

calcloud/submit.py
```diff
@@ -17,10 +17,10 @@ def submit_job(plan_tuple):
         "jobQueue": info.job_queue,
         "jobDefinition": info.job_definition,
         "containerOverrides": {
-            "resourceRequirements": [
-                {"value": f"{info.memory}", "type": "MEMORY"},
-                {"value": f"{info.vcpus}", "type": "VCPU"},
-            ],
+            # "resourceRequirements": [
```
bhayden53, line 20: just a reminder that the `memory` parameter is deprecated and this is the supported way to set these values in the future, and is required for fargate.
https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html

The really, really annoying thing though is that I am pretty sure you cannot see the values set here anywhere in the console

jaytmiller: In theory these should all be going away in this PR (they're commented out. ) The ladder job definitions in Terraform define cpu and memory so they show up nicely on the console starting now.

bhayden53: From the above link:

> memory
This parameter is deprecated and not supported for jobs run on Fargate resources. Use ResourceRequirement instead. For jobs run on EC2 resources that aren't using ResourceRequirement, the number of MiB of memory reserved for the job. For other jobs, see resourceRequirements. If your container attempts to exceed the memory specified here, the container is killed. This parameter maps to Memory in the Create a container section of the Docker Remote API and the --memory option to docker run. You must specify at least 4 MiB of memory for a job. This is required but can be specified in several places for multi-node parallel (MNP) jobs. It must be specified for each node at least once.

jaytmiller: I see.   Fantastic point,  I will fix that!

jaytmiller: OK,  I pushed the fix for ResourceRequirements in batch.tf.

terraform/batch.tf
```diff
@@ -168,6 +173,98 @@ resource "aws_batch_job_definition" "calcloud" {
...
+  container_properties = <<CONTAINER_PROPERTIES
+  {
+    "command": ["Ref::command", "Ref::dataset", "Ref::input_path", "Ref::s3_output_path", "Ref::crds_config"],
+    "environment": [],
+    "image": "${aws_ecr_repository.caldp_ecr.repository_url}:${data.aws_ecr_image.caldp_latest.image_tag}",
+    "jobRoleArn": "${data.aws_ssm_parameter.batch_job_role.value}",
+    "mountPoints": [],
+    "resourceRequirements": [
+    	{"value": "${16*(1024-128)}", "type": "MEMORY"},
+        {"value": "8", "type": "VCPU"}
```
bhayden53, line 220: indent

terraform/batch.tf
```diff
@@ -168,6 +173,98 @@ resource "aws_batch_job_definition" "calcloud" {
...
+  container_properties = <<CONTAINER_PROPERTIES
+  {
+    "command": ["Ref::command", "Ref::dataset", "Ref::input_path", "Ref::s3_output_path", "Ref::crds_config"],
+    "environment": [],
+    "image": "${aws_ecr_repository.caldp_ecr.repository_url}:${data.aws_ecr_image.caldp_latest.image_tag}",
+    "jobRoleArn": "${data.aws_ssm_parameter.batch_job_role.value}",
+    "mountPoints": [],
+    "resourceRequirements": [
+    	{"value": "${64*(1024-128)}", "type": "MEMORY"},
+        {"value": "32", "type": "VCPU"}
```
bhayden53, line 250: indent

bhayden53:
LGTM for as much as I can tell, will run some tests on sandbox

jaytmiller:
OK,  thanks for the detailed review,  nice saves in there!   Merging.

PR #59, opened 2021-03-08
terraform provider and module version pinning

bhayden53:
unfortunately there doesn't seem to be a way to globally pin the `terraform-aws-modules/lambda/aws` version, and you also cannot turn it into a variable (because it is only valid at `init`, it cannot be changed on-the-fly in a `plan` or `apply`). So the only option seems to be to specify it uniquely in every instantiation of an object from the module.

### DIFF
--- a/terraform/batch.tf
+++ b/terraform/batch.tf
@@ -2,6 +2,35 @@ provider "aws" {
   region  = var.region
 }
 
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      version = "~> 3.29.0"
+    }
+    hashicorp-template = {
+      source = "hashicorp/template"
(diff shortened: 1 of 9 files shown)


jaytmiller (approved):
LGTM.   Maybe document  the current version of the Terraform program in .tf or the ticket and then work with Tanner/Leigh to get that version pinned in the CI-node launch template universally.

bhayden53:
> LGTM. Maybe document the current version of the Terraform program in .tf or the ticket and then work with Tanner/Leigh to get that version pinned in the CI-node launch template universally.

I added a variable with the intended pinned terraform version, and confirmed with ITSD how to get changes propagated through the user_data scripts in all of our environments.

PR #60, opened 2021-03-08
Calcloud 223 async inputs

jaytmiller:
Adds a wait loop watching for the input tarball and memory control file to ensure the job is ready to plan and execute.   Extended lambda timeouts for submit and rescue to 15 min / 900 seconds.
Fixed a bug in S3 list pagination and max objects limits.

### DIFF
--- a/calcloud/lambda_submit.py
+++ b/calcloud/lambda_submit.py
@@ -12,6 +12,7 @@ See the batch_event handler, rescue handler, and s3_trigger
 handlers for more information on how jobs are initiated and
 retried.
 """
+import time
 
 from . import plan
 from . import submit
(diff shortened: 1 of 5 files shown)


calcloud/lambda_submit.py
```diff
@@ -58,3 +62,31 @@ def _main(comm, ipppssoot, bucket_name):
...
+    input_tarball, memory_modeling = [], []
+    while not input_tarball or not memory_modeling:
+        input_tarball = comm.inputs.listl(f"{ipppssoot}.tar.gz")
+        memory_modeling = comm.control.listl(f"{ipppssoot}/{ipppssoot}_MemModelFeatures.txt")
+        if not comm.messages.listl([f"placed-{ipppssoot}", f"rescue-{ipppssoot}"]):
+            raise CalcloudInputsFailure(
+                f"Both the 'placed' and 'rescue' messages for {ipppssoot} have been deleted. Aborting input wait and submission."
+            )
+        if not input_tarball or not memory_modeling:
+            print(
```
bhayden53, line 88: might be good to log which one is actually missing

jaytmiller: I think the print on line 88 does that,  but I may be missing something.

bhayden53: oh this snippet was referring to line 88 but the rest of the statement is on 89 and apparently GH snippets the previous 4 lines.

Will this print an empty list or something for the one that's missing? I guess I would have to go see what `listl` does

jaytmiller: Yes,  that's what it does.   list() returns an iterator which has a repr() which isn't good for much.  listl() is wrapper around list() which evaluates it and returns a list which has a meaningful str/repr().

I'm not really happy with those names.   I found myself wishing listl() was just named ls().   I'm also unsure the iterator versions of methods/functions should have normal looking names because they're surprising when you use them outside loops and e.g. I think they evaluate to True regardless of the number of members in the sequence.

bhayden53:
LGTM if you're done

PR #61, opened 2021-03-11
added shorter variable for CSYS_VER due to cataloging constraints

bhayden53:


lisaesherbert:
But in the next sprint the field will increase to 70 characters. See DADS-1777

bhayden53:
> But in the next sprint the field will increase to 70 characters. See DADS-1777

I didn't want it to block other work in the meantime, it's a very simple change, I can always undo it.

bhayden53:
violet is increasing the length of the keyword so there's no need to actually merge this, I can just use it temporarily to keep cataloging working

PR #62, opened 2021-03-12
new refresh cache lambda

bhayden53:
Adds lambda to call refresh cache operation every 5 minutes on the core prefixes, and every 30 or so on the less important ones

### DIFF
--- a/lambda/blackboard/scrape_batch.py
+++ b/lambda/blackboard/scrape_batch.py
@@ -35,7 +35,6 @@ def lambda_handler(event, context):
 
     # we need s3 to upload the snapshot, and storagegateway to refresh the cache
     s3 = boto3.client("s3", config=common.retry_config)
-    gateway = boto3.client("storagegateway", config=common.retry_config)
 
     with os.fdopen(fd, "w") as fout:
         # write the header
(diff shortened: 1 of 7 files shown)


jaytmiller (commented):
I'll need to study some to back you up on this,  but off the cuff:  other than lambda logs,  how do we test this?

bhayden53:
There's no way to test it without a gateway really, unless we completely mock up the gateway API. We could test it by passing it different event times and making sure all of the prefixes get called in the right places?

But in terms of what AWS does behind the scenes, it's either manual, or a total mock-up.

jaytmiller (approved):
This may be irrelevant but checking the lambda CloudWatch in sandbox is showing exceptions in the handler.   Other than that,  LGTM.

bhayden53:
The version in sandbox hasn't actually been running, there was an issue with the scheduler event rule not mapping to multiple lambdas that I fixed in dev but never pushed to sandbox.

The problem on sandbox is that they won't let us have a gateway there, so the exception is because there's a dummy FILESHARE variable in SSM Parameter store, but no actual gateway or file share.

So if you check dev you can see it working. Then you have to go to the log stream of the other lambda to see the captured events... it occurs to me we could create our own single log stream to put both of these into, but I'll think about that later.

PR #63, opened 2021-03-15
CALCLOUD-228 Enhanced CALDP error handling and CALCLOUD failure events

jaytmiller:
Adding a sysexit.py module which supplies the exit_on_exception() and exit_receiver() context managers which support bracketing blocks of code and throwing any exception raised  back up to the program top level.
Adding the exit_codes.py module which defines CALDP error codes and text explanations.
Modifying caldp-process to be more transparent to the error codes issued by process.py and create_previews.py
Modifying process.py, create_previews.py, and messages.py to trap different error cases and report them with unique error codes.
Modifying the batch event handler to handle more event formats,  including Container memory limit, Python MemoryError, and operator cancelled.
Added a search for MemoryError in caldp-process using grep to find trapped memory errors in process.txt
Improved CALCLOUD batch event handler to process varying event formats for: container memory limit, Python memory error, program non-zero exit code,  operator cancelled event.

### DIFF
--- a/calcloud/exit_codes.py
+++ b/calcloud/exit_codes.py
@@ -1,0 +1,141 @@
+"""These numerical values are returned by CALDP as process exit status.
+
+This file should be shared/coordinated *verbatim* with CALCLOUD to ensure that
+CALCLOUD correctly identifies and handles exit status coming from CALDP.
+
+The intent of these codes is to identify specific error cases defined by CALDP.
+
+Any errors not explicitly handled by CALDP are intended to be mapped to
+generic values of 0 or 1 to prevent conflicts with these codes.
+"""
+import sys
+
+_EXIT_CODES = dict(
+    SUCCESS=0,
+    GENERIC_ERROR=1,
+    CMDLINE_ERROR=2,
+    INPUT_TAR_FILE_ERROR=21,
+    ASTROQUERY_ERROR=22,
+    STAGE1_ERROR=23,
+    STAGE2_ERROR=24,
+    S3_UPLOAD_ERROR=25,
+    S3_DOWNLOAD_ERROR=26,
+    BESTREFS_ERROR=27,
+    CREATE_PREVIEWS_ERROR=28,
+    SUBPROCESS_MEMORY_ERROR=31,  # See caldp-process for this
+    CALDP_MEMORY_ERROR=32,
+    CONTAINER_MEMORY_ERROR=33,
+)
+
+
+_NAME_EXPLANATIONS = dict(
+    SUCCESS="Processing completed successfully.",
+    GENERIC_ERROR="An error with no specific CALDP handling occurred somewhere.",
+    CMDLINE_ERROR="The program command line invocation was incorrect.",
+    INPUT_TAR_FILE_ERROR="An error occurred locating or untarring the inputs tarball.",
+    ASTROQUERY_ERROR="An error occurred downloading astroqery: inputs",
+    STAGE1_ERROR="An error occurred in this instrument's stage1 processing step. e.g. calxxx",
+    STAGE2_ERROR="An error occurred in this instrument's stage2 processing step, e.g astrodrizzle",
+    S3_UPLOAD_ERROR="An error occurred uploading the outputs tarball to S3.",
+    S3_DOWNLOAD_ERROR="An error occurred downloading inputs from S3.",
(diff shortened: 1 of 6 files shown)


calcloud/exit_codes.py
```diff
@@ -0,0 +1,141 @@
...
+    >>> is_memory_error(CONTAINER_MEMORY_ERROR)
+    True
+
+    >>> is_memory_error("GENERIC_ERROR")
+    False
+
+    >>> is_memory_error("SUBPROCESS_MEMORY_ERROR")
+    True
+    """
+    memory_error_names = ["SUBPROCESS_MEMORY_ERROR", "CALDP_MEMORY_ERROR", "CONTAINER_MEMORY_ERROR"]
```
bhayden53, line 120: if you ever change _NAME_EXPLANATIONS do you need to change this list as well? Might be hard to remember. Maybe a separate global?

lambda/batch_events/batch_event_handler.py
```diff
@@ -20,21 +20,24 @@
...
+    job_id = detail["jobId"]
+    job_name = detail["jobName"]  # appears to be ipppssoot
+    status_reason = detail.get("statusReason", "undefined")
+
+    container = event["detail"]["container"]
+    ipppssoot = container["command"][1]
+    bucket = container["command"][2].split("/")[2]
+    container_reason = container.get("reason", "undefined")
+    exit_code = container.get("exitCode", "undefined")
+    exit_reason = exit_codes.explain(exit_code) if exit_code != "undefined" else exit_code
```
bhayden53, line 40: why not let exit_codes.explain handle "undefined" ?

jaytmiller: I thought about that but in the end refrained because "undefined" is  neither an integer code nor a name for a code and this is only used in one place.

terraform/lambda_job_submit.tf
```diff
@@ -38,7 +38,7 @@ module "calcloud_lambda_submit" {
   attach_tracing_policy = false
   attach_async_event_policy = false
   # existing role for the lambda
-  # will need to parametrize when ITSD takes over role creation. 
+  # will need to parametrize when ITSD takes over role creation.
```
bhayden53, line 41: not sure why this is showing up in the diff but it makes me realize this comment is way deprecated

calcloud/exit_codes.py
jaytmiller: I converted this to _MEMORY_ERROR_NAMES at the top of the file.

terraform/lambda_job_rescue.tf
```diff
@@ -37,10 +37,10 @@ module "calcloud_lambda_rescueJob" {
   attach_tracing_policy = false
   attach_async_event_policy = false
   # existing role for the lambda
-  # will need to parametrize when ITSD takes over role creation. 
+  # will need to parametrize when ITSD takes over role creation.
```
bhayden53, line 40: this one deprecated too

calcloud/exit_codes.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    if exit_code in _CODE_TO_NAME:
+        name = _CODE_TO_NAME[exit_code]
+        explanation = _NAME_EXPLANATIONS[name]
+    elif exit_code in _NAME_EXPLANATIONS:
+        name = exit_code
+        exit_code = globals()[name]
+        explanation = _NAME_EXPLANATIONS[name]
+    else:
+        raise ValueError("Invalid exit_code: " + repr(exit_code))
```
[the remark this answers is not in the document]
jaytmiller: I changed this to Unhandled.

bhayden53:
LGTM, remaining comments are optional

terraform/lambda_job_submit.tf
jaytmiller: I added whitespace handling to me .emacs config so it automatically trimmed trailing whitespace.   You can see it  as the dark red block at the end of the old comment.

terraform/lambda_job_rescue.tf
jaytmiller: Same whitespace issue.

terraform/lambda_job_submit.tf
bhayden53: yeah the side-by-side diff didn't show the red block, it only showed up in the conversation-thread diff

PR #64, opened 2021-03-16
example of an environment-specific variable and generic default

bhayden53:
This is a demonstration of how to setup environment-specific terraform variables.

In this case, use a map(string) variable with environment keys to determine the right EBS type (gp2 for lower-load envs, gp3 for higher load). Use the lookup function to default to a generic "gp2" if the environment is not one of the "official" ones, (i.e., if it's "jmiller")


jaytmiller (approved):
LGTM

PR #65, opened 2021-03-17
Clean(up) lambda

jaytmiller:
Added clean-xxx lambda to wipe out all S3 files for an ipppssoot or all.
Added broadcast-kill safety feature to force exits in runaway broadcasts.
Added input checking to MessageIo.broadcast()
Additional work on clean(), reset(), and ids() methods.  Ids are effectively
  ipppssoots vs full filenames or messages.

### DIFF
--- a/terraform/lambda_job_clean.tf
+++ b/terraform/lambda_job_clean.tf
@@ -1,0 +1,60 @@
+module "calcloud_lambda_cleanJob" {
+  source = "terraform-aws-modules/lambda/aws"
+  version = "~> 1.43.0"
+
+  function_name = "calcloud-job-clean${local.environment}"
+  description   = "accepts messages from s3 event and cleans either individual jobs by ipppssoot, or all active jobs"
+  # the path is relative to the path inside the lambda env, not in the local filesystem.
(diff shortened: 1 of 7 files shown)


terraform/lambda_job_clean.tf
```diff
@@ -0,0 +1,60 @@
...
+  create_role = false
+  attach_cloudwatch_logs_policy = false
+  attach_dead_letter_policy = false
+  attach_network_policy = false
+  attach_tracing_policy = false
+  attach_async_event_policy = false
+  # existing role for the lambda
+  # will need to parametrize when ITSD takes over role creation.
+  # for now this role was created by hand in the console, it is not terraform managed
+  lambda_role = data.aws_ssm_parameter.lambda_delete_role.value
```
bhayden53, line 42: only concern I have with re-using the delete role is that the clean lambda doesn't "need" Batch perms.

jaytmiller: You're right,  thanks for catching this!   I worked Leigh and we set up a cleanup lambda role.

bhayden53:
this has my approval for merge, if you're done

PR #67, opened 2021-03-20
LambdaJobPredict

alphasentaurii:
Lambda function from docker image for predicting job resource allocation requirements


calcloud/plan.py
```diff
@@ -16,6 +16,11 @@
 from . import log
 from . import s3
 
+import json
+import boto3
+
+client = boto3.client("lambda")
```
bhayden53, line 22: there's a config object in the "common" module that we have been using in all clients to config retries.

lambda/JobPredict/predict_handler.py
```diff
@@ -0,0 +1,177 @@
...
+def regressor(model, data):
+    """Returns Regression model prediction"""
+    pred = model.predict(data)
+    return pred
+
+
+clf = get_model("./models/mem_clf/1/")
+mem_reg = get_model("./models/mem_reg/1/")
+wall_reg = get_model("./models/wall_reg/1/")
+s3 = boto3.resource("s3")
```
bhayden53, line 143: potentially another place to use a custom retry config.

I think the default is 2 retries; I like to up it to 5, as we don't have a lot built-in to recover from outright lambda failures. Our fallback is just resending the placed- message or even re-submitting the repro on-prem.

lambda/JobPredict/predict_handler.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def lambda_handler(event, context):
+    """Predict Resource Allocation requirements for memory (GB) and max execution `kill time` / `wallclock` (seconds) using three pre-trained neural networks. This lambda is invoked from the Job Submit lambda which json.dumps the s3 bucket and key to the file containing job input parameters. The path to the text file in s3 assumes the following format: `control/ipppssoot/ipppssoot_MemModelFeatures.txt`.
+
+    MEMORY BIN: classifier predicts which of 4 memory bins is most likely to be needed to process an HST dataset (ipppssoot) successfully. The probabilities of each bin are output to Cloudwatch logs and the highest bin probability is returned to the Calcloud job submit lambda invoking this one. Bin sizes are as follows:
+
+    Memory Bins:
```
bhayden53, line 151: Because of EC2 memory allocations and overhead issues, Todd had to reduce the max memory of each bin a tiny bit; see i.e
https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/terraform/batch.tf#L170

Is your prediction allowing some overhead? Do we need to re-train with Todd's limits instead of the nice round numbers?

setup.cfg
```diff
@@ -32,6 +32,6 @@ dev =
 [flake8]
 ignore = E501, E203, W503
 exclude = .git, __pycache__, build, dist, eggs, *.egg
-
+per-file-ignores = ./lambda/JobPredict/predict_handler.py:F401
```
bhayden53, line 35: "module imported but unused" ?

terraform/lambda_job_submit.tf
```diff
@@ -45,6 +45,8 @@ module "calcloud_lambda_submit" {
   environment_variables = {
     JOBDEFINITIONS = local.job_definitions,
     NORMALQUEUE = aws_batch_job_queue.batch_queue.name,
+    S3BUCKET = aws_s3_bucket.calcloud.id,
```
bhayden53, line 48: the s3 bucket is actually read from the s3 event and passed into calcloud through lambda_submit.main.

https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/s3_trigger/s3_trigger_handler.py#L8

https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/s3_trigger/s3_trigger_handler.py#L16

We used to have the s3 bucket as an env variable here but it just felt like better design to pull it straight from the event. If there's an easy way for you to get a hold of that version of the variable in plan.py, that would be preferred.

terraform/locals.tf
```diff
@@ -10,4 +10,7 @@ locals {
        environment = local.pre_environment == "" ? "" : join("", ["-", local.pre_environment])
 
        job_definitions = "${aws_batch_job_definition.calcloud_2g.name},${aws_batch_job_definition.calcloud_8g.name},${aws_batch_job_definition.calcloud_16g.name},${aws_batch_job_definition.calcloud_64g.name}"
+
+       ecr_address = format("%v.dkr.ecr.%v.amazonaws.com", data.aws_caller_identity.this.account_id, data.aws_region.current.name)
+       ecr_image   = format("%v/%v:model", local.ecr_address, aws_ecr_repository.caldp_ecr.name)
```
bhayden53, line 15: since we'll have multiple images and this is a somewhat global terraform file, could you rename this to be more specific about which ecr_image it refers to?

lambda/JobPredict/predict_handler.py
```diff
@@ -0,0 +1,177 @@
...
+    0: < 2GB
+    1: 2-8GB
+    2: 8-16GB
+    3: >16GB
+
+    WALLCLOCK REGRESSION: regression generates estimate for specific number of seconds needed to process the dataset using the same input data. This number is then tripled in Calcloud for the sake of creating an extra buffer of overhead in order to prevent larger jobs from being killed unnecessarily.
+
+    MEMORY REGRESSION: A third regression model is used to estimate the actual value of memory needed for the job. This is mainly for the purpose of logging/future analysis and is not currently being used for allocating memory in calcloud jobs.
+    """
+    bucket_name = event["Bucket"]
```
jaytmiller, line 161: It's minor,  but since the lambda is exposed as a function,  it may need parameter screening as a security thing.   

Potentially the bucket and key could be defined in the environment to avoid the need to screen, and ipppssoot can be screened  with:

     assert hst.IPPPSSOOT_RE.match(ipppssoot)

This is not needed for other lambdas called from S3 events.

jaytmiller (approved):
LGTM,  only comment which is important is the possible bug.

bhayden53:
your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36

That may all be IAM? The problem is that rescue re-uses the delete role, and the delete lambda doesn't need invoke permissions, so that actually implies we need a totally new rescue IAM role with the additional invoke permissions

terraform/lambda_job_submit.tf
alphasentaurii: Yes, in plan.py the "output_bucket" var is "s3://bucket" so I can just strip the "s3://" prefix and pass that through instead

setup.cfg
alphasentaurii: terraform and keras are not installed until the container is built from image so flake8 throws an error: 
```python
./lambda/JobPredict/predict_handler.py:9:1: F401 'sklearn' imported but unused
./lambda/JobPredict/predict_handler.py:11:1: F401 'tensorflow as tf' imported but unused
./lambda/JobPredict/predict_handler.py:13:1: F401 'tensorflow.keras.models.load_model' imported but unused
./lambda/JobPredict/predict_handler.py:14:1: F401 'tensorflow.keras.models.Sequential' imported but unused
./lambda/JobPredict/predict_handler.py:15:1: F401 'tensorflow.keras.layers.Dense' imported but unused 
```

bhayden53: But using load_model as the example, you don't use the imported one, you pull it right out of keras, so flake8 is technically correct:
`model = keras.models.load_model(model_path)`

terraform/locals.tf
alphasentaurii: `ecr_predict_lambda_address`, `ecr_predict_lambda_image` ?

lambda/JobPredict/predict_handler.py
```diff
@@ -0,0 +1,177 @@
...
+        pctecorr = 0
+        crsplit = 0
+
+        for k, v in self.input_data.items():
+            if k == "n_files":
+                n_files = int(v)
+            if k == "total_mb":
+                total_mb = int(np.round(float(v), 0))
+            if k == "DETECTOR":
+                if v == "UVIS" or "WFC":
```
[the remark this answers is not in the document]
alphasentaurii: corrected - good catch

terraform/locals.tf
bhayden53: I actually don't think it matters for ecr_address, I was only referring to ecr_image, but github shows the extra lines around the comment...

`ecr_predict_lambda_image` is fine

lambda/JobPredict/predict_handler.py
alphasentaurii: hst.py (along with other calcloud python files) are not included in the container image for the predict lambda but I can pass it through the event that invokes it (i.e. from plan.py). you're suggesting I get the Bucket and Key as environment variables instead of through the event invocation?

bhayden53:
> > your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
> > https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36
> > 
> > That may all be IAM? The problem is that rescue re-uses the delete role, and the delete lambda doesn't need invoke permissions, so that actually implies we need a totally new rescue IAM role with the additional invoke permissions
> 
> @bhayden53 Can you clarify what you mean here? In which scenario would the rescue lambda need to invoke the memory model? My understanding was that it would be using the original prediction vals (possibly being stored in the `comm` var or in s3/control in order to see which bin was used and iterate to the next highest one.

When we talked about it on the call we decided not to put the initial memory bin into the control file because it was too complex; we decided that both lambdas would need to invoke your model lambda to get the initial memory bin.

The rescue lambda directly calls `lambda_submit.main()`, which eventually ends up in `plan._get_resources`, which after your changes, invokes your new lambda. If merged, this PR in current form will cause every rescue to fail because the role used by the rescue lambda cannot invoke your function.

lambda/JobPredict/predict_handler.py
alphasentaurii: No it is not - model predicts bin values according to the original ranges we discussed:

- <2 GB
- >=2.0 and <8.0
- >=8.0 and < 16.0
- >=16

So yes I would need to re-classify the data according to the modified ranges and re-train. Shouldn't be a problem...

bhayden53:
> > > your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
> > > https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36
> > > 
> > > That may all be IAM? The problem is that rescue re-uses the delete role, and the delete lambda doesn't need invoke permissions, so that actually implies we need a totally new rescue IAM role with the additional invoke permissions
> > 
> > 
> > @bhayden53 Can you clarify what you mean here? In which scenario would the rescue lambda need to invoke the memory model? My understanding was that it would be using the original prediction vals (possibly being stored in the `comm` var or in s3/control in order to see which bin was used and iterate to the next highest one.
> 
> When we talked about it on the call we decided not to put the initial memory bin into the control file because it was too complex; we decided that both lambdas would need to invoke your model lambda to get the initial memory bin.
> 
> The rescue lambda directly calls `lambda_submit.main()`, which eventually ends up in `plan._get_resources`, which after your changes, invokes your new lambda. If merged, this PR in current form will cause every rescue to fail because the role used by the rescue lambda cannot invoke your function.

you can try this out to be sure by posting a `rescue-ipppssoot` message in your sandbox environment and checking the rescue lambda logs.

calcloud/plan.py
alphasentaurii: like so?
`client = boto3.client("lambda", config=common.retry_config)`

bhayden53: yep!

terraform/lambda_job_submit.tf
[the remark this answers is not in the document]
alphasentaurii: May pass bucket through event instead of environment after all (see Brian's comment) but good to know

alphasentaurii:
> > > your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
> > > https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36
> > > 
> > > That may all be IAM? The problem is that rescue re-uses the delete role, and the delete lambda doesn't need invoke permissions, so that actually implies we need a totally new rescue IAM role with the additional invoke permissions
> > 
> > 
> > @bhayden53 Can you clarify what you mean here? In which scenario would the rescue lambda need to invoke the memory model? My understanding was that it would be using the original prediction vals (possibly being stored in the `comm` var or in s3/control in order to see which bin was used and iterate to the next highest one.
> 
> When we talked about it on the call we decided not to put the initial memory bin into the control file because it was too complex; we decided that both lambdas would need to invoke your model lambda to get the initial memory bin.
> 
> The rescue lambda directly calls `lambda_submit.main()`, which eventually ends up in `plan._get_resources`, which after your changes, invokes your new lambda. If merged, this PR in current form will cause every rescue to fail because the role used by the rescue lambda cannot invoke your function.

Got it - yes that's purely IAM change since rescue calls plan to invoke lambda, no code changes should be necessary.

alphasentaurii:
- [x] rescue lambda IAM permissions (Invoke JobPredict lambda)
- [x] terraform/locals.tf: ecr_predict_lambda_image
- [x] lambda/JobPredict/predict_handler.py:
          - [x] retry_config
          - [x] line 51-52: if v in ["UVIS", "WFC”]: 
          - [x] line 161: pass ipppssoot through event 
          - [x] remove unused module imports
- [x] lambda_job_submit.tf: strip "s3://" prefix, pass through event instead of environment
- [x] plan.py: retry_config; pass ipppssoot through event also
- [x] setup.cfg: remove flake8 rule ignoring unused imports
- [x] testing: job submits and rescue lambda

bhayden53:
> * [x]  rescue lambda IAM permissions (Invoke JobPredict lambda)
> * [x]  terraform/locals.tf: ecr_predict_lambda_image
> * [x]  lambda/JobPredict/predict_handler.py:
>   - [x] retry_config
>   - [x] line 51-52: if v in ["UVIS", "WFC”]:
>   - [x] line 161: pass ipppssoot through event
>   - [x] remove unused module imports
> * [x]  lambda_job_submit.tf: strip "s3://" prefix, pass through event instead of environment
> * [x]  plan.py: retry_config; pass ipppssoot through event also
> * [x]  setup.cfg: remove flake8 rule ignoring unused imports
> * [x]  testing: job submits and rescue lambda



> > > > your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
> > > > https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36
> > > > 
> > > > That may all be IAM? The problem is that rescue re-uses the delete role, and the delete lambda doesn't need invoke permissions, so that actually implies we need a totally new rescue IAM role with the additional invoke permissions
> > > 
> > > 
> > > @bhayden53 Can you clarify what you mean here? In which scenario would the rescue lambda need to invoke the memory model? My understanding was that it would be using the original prediction vals (possibly being stored in the `comm` var or in s3/control in order to see which bin was used and iterate to the next highest one.
> > 
> > 
> > When we talked about it on the call we decided not to put the initial memory bin into the control file because it was too complex; we decided that both lambdas would need to invoke your model lambda to get the initial memory bin.
> > The rescue lambda directly calls `lambda_submit.main()`, which eventually ends up in `plan._get_resources`, which after your changes, invokes your new lambda. If merged, this PR in current form will cause every rescue to fail because the role used by the rescue lambda cannot invoke your function.
> 
> Got it - yes that's purely IAM change since rescue calls plan to invoke lambda, no code changes should be necessary.

It's not quite purely IAM, as we'll need to change the terraform for rescue to reference the new IAM role. I think that's all that's left before merge.

alphasentaurii:
> > > > > your lambda will also need permission to be called from the rescue lambda because it can directly submit jobs:
> > > > > https://github.com/spacetelescope/calcloud/blob/c83cf29066e403b2e7c4d93e04411f0e172070f2/lambda/JobRescue/rescue_handler.py#L36

> It's not quite purely IAM, as we'll need to change the terraform for rescue to reference the new IAM role. I think that's all that's left before merge.

@bhayden53 New role has been created, updating rescue lambda and testing now.

PR #72, opened 2021-03-24
increment calcloud version in prep for release, and update base image

bhayden53:


bhayden53:
can't merge base image because the proper CRDS contexts haven't hit Ops yet and we have no access to test contexts

PR #75, opened 2021-03-25
CRDS config for using the official S3 bucket

bhayden53:
* CRDS on S3 and Context Env Variables passed around through terraform
* quick fix for short wallclock times
* updating base image
* goes with caldp PR https://github.com/spacetelescope/caldp/pull/57

### DIFF
--- a/terraform/locals.tf
+++ b/terraform/locals.tf
@@ -14,8 +14,12 @@ locals {
        ecr_address = format("%v.dkr.ecr.%v.amazonaws.com", data.aws_caller_identity.this.account_id, data.aws_region.current.name)
        ecr_predict_lambda_image   = format("%v/%v:model", local.ecr_address, aws_ecr_repository.caldp_ecr.name)
 
-       # if bucket == test, return test ssm param, elif bucket == ops, return ops ssm param, else return test ssm param
-       crds_bucket = lookup(var.crds_bucket, local.environment, "test") == "test" ? data.aws_ssm_parameter.crds_test.value : (lookup(var.crds_bucket, local.environment, "test") == "ops" ? data.aws_ssm_parameter.crds_ops.value : data.aws_ssm_parameter.crds_test.value)
+       # because we cannot reference ssm params in variables, we have to set the crds bucket here by looking up the desired bucket through a string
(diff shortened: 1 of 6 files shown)


terraform/locals.tf
```diff
@@ -13,4 +13,10 @@ locals {
 
        ecr_address = format("%v.dkr.ecr.%v.amazonaws.com", data.aws_caller_identity.this.account_id, data.aws_region.current.name)
        ecr_predict_lambda_image   = format("%v/%v:model", local.ecr_address, aws_ecr_repository.caldp_ecr.name)
+
+       # if bucket == test, return test ssm param, elif bucket == ops, return ops ssm param, else return test ssm param
+       crds_bucket = lookup(var.crds_bucket, local.environment, "test") == "test" ? data.aws_ssm_parameter.crds_test.value : (lookup(var.crds_bucket, local.environment, "test") == "ops" ? data.aws_ssm_parameter.crds_ops.value : data.aws_ssm_parameter.crds_test.value)
```
jaytmiller, line 18: Something like this may work better if this comes up a lot: 

crds_bucket = defaults({
   "test" : data.aws_ssm_parameter.crds_test.value,
   "ops" : data.aws_ssm_parameter.crds_ops.value,
},  data.aws_ssm_parameter.crds_test.value)[local.environment]

It's untested code but I suspect it works,  the key being (a) use of a map and (b) creating a default for the map by calling defaults().

jaytmiller (approved):
LGTM.

terraform/locals.tf
bhayden53: Thanks, I had tried to use a map in variables.tf, but forgot that you can't reference any ssm params there; so then I moved this logic into batch.tf, but it was messy, so I moved it back out to locals. it hadn't occurred to me to make it a map in locals!

I think your snippet is missing a lookup in var.crds_bucket to determine what default crds to use in each of -sb, -dev, -test, -ops, but I get the gist of it. May not change it now for lack of time.

bhayden53: i.e. local.environment should be lookup(var.crds_bucket, local.environment, "test")

bhayden53: pushed a change, thanks for the map suggestion, looks way cleaner and is easier to understand

PR #77, opened 2021-03-30
Pin lambda_job_predict version

jaytmiller:
Pinned lambda_job_predict module version back to 1.43.0 same as all other lambdas.  This is an untested regression from 1.44.0.


bhayden53 (approved):
lgtm

alphasentaurii (approved):

PR #78, opened 2021-03-30
actually use the max cpu var and fix crds contexts

bhayden53:
* change max cpu in test to 512 and actually use it in the compute environment
* updates to default crds contexts

### DIFF
--- a/terraform/batch.tf
+++ b/terraform/batch.tf
@@ -125,7 +125,7 @@ resource "aws_batch_compute_environment" "calcloud" {
     subnets             = local.batch_subnet_ids
     security_group_ids  = local.batch_sgs
     instance_type = ["optimal"]
-    max_vcpus = 128
+    max_vcpus = lookup(var.ce_max_vcpu, local.environment, 64)
     min_vcpus = 0
     desired_vcpus = 0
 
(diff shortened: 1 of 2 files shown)


jaytmiller (approved):
I was wondering if the optimal ec2 work needed more variables yesterday and looking at this see a few potential issues (a) there will be a minor coding conflict with ce_max_vcpus,  but it does resolve one TBD I left in locals.tf (b) I'm wondering about the benefits of using a variable for ce_max_vcpus vs. just hard coding in locals.tf.  any preference?  (c) I'm wondering how lookups from the defaults work with non-standard environments like -sb-jmiller.  we could partition environment into two pieces to make that work if it does not already:  environment = "-sb" (account) + "-jmiller"  or "" (deployment).

bhayden53:
The lookup will fall back on the third parameter, which is the value to use if not found. i.e. `lookup(cpus, "-sb-jmiller", 64)` would use 64 if it couldn't find your env in the map. So lookup does pretty much all of the work in handling non-standard env names. I am pretty much fine with that because there shouldn't be any non-standard envs outside of sandbox anyway.

I guess I put the CPUs setting in locals because anything that I wanted to spread out to different values in different envs made more sense to me there. But I didn't think about it much deeper than that.

PR #79, opened 2021-03-31
Calcloud 236 optimal ec2s

jaytmiller:
This is the general gist of the ladder upgrade with bin-specific queues and instances.  I haven't run it yet,  but it does deploy and I'm checking it over.


jaytmiller:
This PR is now complete,  I ran one data set through the 8G bin and it seemed to work fine.

Along the way I discovered that the 6g series instances are actuallly ARM,  whereas the 5 series  is  x86.

bhayden53 (approved):
lgtm

PR #80, opened 2021-04-01
CRDS context pinning via crds_env_vars path in primary S3 bucket

bhayden53:
companion to https://github.com/spacetelescope/caldp/pull/59

will squash merge


bhayden53:
this is almost certainly not going to work. When I update the job definition in this way, terraform loses track, because terraform tracks a specific revision number. So the next time terraform goes to apply, it sees that it's jobdefs are just gone, and makes totally new ones. So then you end up with 2 sets of jobdefs in ACTIVE state.

This implies we'll need to pin the CRDS context somewhere else to keep terraform clean.

bhayden53:
refactored to remove lambda implementation completely. It will be up to Operations staff to set the CRDS context correctly.

bhayden53:
ready for review

terraform/deploy.sh
```diff
@@ -77,11 +81,17 @@ bucket_url=${bucket_url_response##*=}
 # removes double quotes from variable
 bucket_url=`echo $bucket_url | tr -d '"'`
 
+# get the crds context
+crds_response=`awsudo $ADMIN_ARN terraform output | grep "crds"`
+crds_context=${crds_response##*=}
+crds_context=`echo $crds_context | tr -d '"'`
+
 awsudo $ADMIN_ARN aws s3api put-object --bucket $bucket_url --key messages/
 awsudo $ADMIN_ARN aws s3api put-object --bucket $bucket_url --key inputs/
```
jaytmiller, line 90: What do these do?   Look like they could create objects vs. directories since (I think) key=outputs/ includes the / and  is actually flat.  Thought the directory tree structure on the  console was  an illusion.  What is body? ""?  crds_context makes  sense though.

jaytmiller (approved):
One Q,  but LGTM.

terraform/deploy.sh
bhayden53: In the past, the on-premise system has crashed because the folders did not exist in the on-premise mount point. Adding these to the end of the deploy process ensures that they're created if the S3 bucket is being stood up from scratch, so that the on-premise codebase doesn't need to do any "does the directory exist yet" shenanigans, which really should be done on the cloud side as part of a normal deploy.

I pulled the solution from here:
https://forums.aws.amazon.com/thread.jspa?messageID=961724

Where one of the comments is that by using a trailing slash it ensures that a folder is created. Whatever the case may be, this does ensure that the directories show up on-premise, so it seems to work as intended.

PR #84, opened 2021-04-01
bugfix/blackboard jobDuration

alphasentaurii:
completionDate changed from "startedAt" to "stoppedAt". JobDuration was getting zeros because it was subtracting the start time from the start time instead of the end time from the start time.


bhayden53:
are the model files supposed to be in here or did they sneak in accidentally?

jaytmiller (approved):
LGTM

PR #89, opened 2021-04-06
Implement rescue for CannotInspectContainer

jaytmiller:
Terraforms and tox's,  otherwise untested.


bhayden53:
acknowledged lack of testing, I accept the risks; hopefully we will see an error in dev and be able to evaluate if it's working.

PR #93, opened 2021-04-07
Blackboard duration fix

bhayden53:
sets jobDuration to 0 if the job hasn't completed, otherwise the jobDuration shows up as -50 years (in seconds) for jobs that have started but not finished.

### DIFF
--- a/lambda/blackboard/scrape_batch.py
+++ b/lambda/blackboard/scrape_batch.py
@@ -59,7 +59,7 @@ def lambda_handler(event, context):
 
                         # if the job hasn't completed yet, set duration to 0 so it's not -50 years
                         # we check for the stoppedAt attribute, and default to startDate
-                        durationCheck = int(j.get("stoppedAt", jobStartDate) / 1000.0)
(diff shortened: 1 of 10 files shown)


lambda/blackboard/scrape_batch.py
```diff
@@ -56,7 +57,12 @@ def lambda_handler(event, context):
                         jobStartDate = int(j.get("startedAt", default_timestamp) / 1000.0)
                         completionDate = int(j.get("stoppedAt", default_timestamp) / 1000.0)
 
-                        jobDuration = int(completionDate - jobStartDate)
+                        # if the job hasn't completed yet, set duration to 0 so it's not -50 years
+                        # we check for the stoppedAt attribute, and default to startDate
+                        durationCheck = int(j.get("stoppedAt", jobStartDate) / 1000.0)
```
bhayden53, line 62: this is bugged if jobStartDate != 0 actually

bhayden53: the reason I'm jumping through these hoops is that I don't know if I can always trust an == with these floats

alphasentaurii (approved):

PR #98, opened 2021-04-12
calcloud-240 canceled jobs send error message

jaytmiller:
Refactors clean-all lambda and adds "rejection" logic to prevent the batch fail events from 
being processed on all strings.

Extensive refactor of clean lambda to cancel-all based on killable job ids
only, no longer based on ipppssoot control file.  This avoids competing job
kills between job_id-based (required to kill orphan jobs) and
ipppssoot-based (mentioned in control file) and also preserves completed
jobs.

Refactored lambda environments to largely use local.common_env_vars ensuring that BUCKET
is defined in all lambdas.

Added "rejection logic" to kill the batch event handler when the bucket obtained from
the failed job does not match the BUCKET the lambda  was built with.  This stops the
cross-environment event contamination where a secondary environment was processing the
same failure events as the development environment resulting in extraneous error
messages.

Improve signature and documentation for S3Io.put and MessageIo.put, adding the
simple/natural forms S3Io.put(msg_type, payload) and MessageIo.put(msg_type,
payload).  Make signatures of super and subclass put() comply with Liskov
Substitution Principle.

Added payloads to error/terminate messages to indicate which lambda they came from.

### DIFF
--- a/calcloud/batch.py
+++ b/calcloud/batch.py
@@ -78,6 +78,23 @@ def _list_jobs_iterator(queue, status, PageSize=10, client=None):
     return paginator.paginate(jobQueue=queue, jobStatus=status, PaginationConfig={"PageSize": PageSize})
 
 
+def describe_job(job_id, client=None):
+    """Return the description from describe_jobs() for `job_id` or None."""
+    client = client or get_default_client()
+    response = client.describe_jobs(jobs=[job_id])
+    jobs = response["jobs"]
+    if len(jobs):
+        return jobs[0]
+    else:
+        return
+
+
+def get_job_name(job_id, client=None):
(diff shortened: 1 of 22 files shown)


alphasentaurii (commented):
Was deleting remote_state.tf intentional? This tells terraform to initialize using a backend remote state instead of local.

jaytmiller:
Nope,  not intentional,  I'll add it back

jaytmiller:
Remote state is back.  Thanks!

alphasentaurii (approved):
Works as expected on my end (just completed testing in sandbox for cancel-ipst and cancel-all).

bhayden53 (approved):

PR #100, opened 2021-04-13
CC249/refresh-cache-every-5min

alphasentaurii:
CALCLOUD-249: refresh each s3 prefix every time refresh_cache_submit lambda is invoked by EventBridge (set to every 5 mins)

### DIFF
--- a/lambda/refreshCacheSubmit/refresh_cache_submit.py
+++ b/lambda/refreshCacheSubmit/refresh_cache_submit.py
@@ -1,23 +1,15 @@
 def lambda_handler(event, context):
     from calcloud import common
-    import dateutil.parser
     import boto3
     import os
 
     gateway = boto3.client("storagegateway", config=common.retry_config)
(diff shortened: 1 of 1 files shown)


lambda/refreshCacheSubmit/refresh_cache_submit.py
```diff
@@ -1,23 +1,15 @@
...
-    import dateutil.parser
     import boto3
     import os
 
     gateway = boto3.client("storagegateway", config=common.retry_config)
     # we only get two concurrent refresh cache operations, but we'll use them both to get some parallelization
-    # we want to refresh these every time
-    refresh_1 = ["/messages/"]
-    refresh_2 = ["/outputs/", "/blackboard/"]
+    # refresh every 5 minutes via EventBridge
```
bhayden53, line 8: blackboard and crds_env_vars are single files, while control and inputs are huge. Can you move control over to refresh_2 and move one of the small ones up to refresh_1?

PR #104, opened 2021-04-16
model/wallclock-regression-v3

alphasentaurii:
updated weights for wallclock regression model (v3). Model was retrained using dataset from latest batch run on 4-15-2021.


bhayden53:
model files need added

PR #114, opened 2021-05-05
file share / cache refresh refactor

bhayden53:
* the file shares have been split out to one per prefix in the S3 bucket, for performance needs. 
* the cache refresh submit lambda was refactored to account for this change


lambda/refreshCacheSubmit/refresh_cache_submit.py
```diff
@@ -2,25 +2,45 @@ def lambda_handler(event, context):
...
+    dt = dateutil.parser.isoparse(event_time)
 
-    try:
-        response = gateway.refresh_cache(FileShareARN=os.environ["FILESHARE"], FolderList=refresh_2, Recursive=True)
-        print(response)
-    except Exception as exc:
-        print(f"refresh cache failed for {refresh_2} with exception")
-        print(str(exc))
+    # run every time
+    rapid_fileshares = {
```
jaytmiller, line 15: Make sure these are being processed in a reasonable and consistent order.   Historically iteration through dictionary keys was "random" and you still need >= 3.7 to really count on consistent order.

lambda/refreshCacheSubmit/refresh_cache_submit.py
```diff
@@ -2,25 +2,45 @@ def lambda_handler(event, context):
...
+        "outputs": os.environ["FS_OUTPUTS"],
+    }
+
+    # ~once per hour
+    # inputs is never written from the cloud
+    # the only file someone may want quickly on-prem is the memModel features,
+    # but that one is written on-prem so doesn't need a refresh to be visible
+    infrequent_fileshares = {"inputs": os.environ["FS_INPUTS"], "control": os.environ["FS_CONTROL"]}
+
+    for fs_name in rapid_fileshares.keys():
```
jaytmiller, line 28: The AWS gateway docs are saying `Throttle limit: This API is asynchronous, so the gateway will accept no more than two refreshes at any time.`   It may just be loose language,  but it's not saying "per file share".  So w/o more knowledge it's plausible you can overrun the gateway with one lambda because the API call has just initiated or Q'ed a refresh,  not completed one.  Maybe check that.

jaytmiller (approved):
Couple comments,  otherwise LGTM

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: This is an interesting point; FWIW this has been running in Dev for ~20 hours and all 4-6 refresh calls always get through. When I did my deep dive on how all this worked, is when I made up my mind that it's per file share. Now, I can't find exactly where that is explicitly stated. It may have even been an AWS developer forum post. 

This is somewhat backed up by the fact that you make the API call against the file share ARN, and to refresh the cache in the console, you also do it from the file share; but I agree that the AWS language does say "the gateway will accept no more than two". 

This will be something to watch more closely in test, when we send some thousands of jobs through instead just a few dozen.

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: I was willing to accept the "random" order, although I'm willing to switch to an OrderedDict if you think there's a strong reason to enforce consistent ordering.

lambda/refreshCacheSubmit/refresh_cache_submit.py
jaytmiller: Watch for `If you invoke the RefreshCache API when two requests are already being processed, any new request will cause an InvalidGatewayRequestException error because too many requests were sent to the server.` which boto3 docs say can occur.   As far as exception trapping goes in general,  it occurs to me that where possible we should probably avoid failure traps,  or re-raise them,  so we see the failures in the lambda plots.   (Here it could be done by setting a global "exception occurred" flag and then raising an exception at the end if it's True.   Something to consider broadly anyway,  might be nice to see red dots standing out.)

lambda/refreshCacheSubmit/refresh_cache_submit.py
jaytmiller: Sure.   But in particular when we're debugging something,  determinism helps avoid Heisen-bugs due to unpredictable changes in order.    Not a big deal unless you're getting flaky behavior,  then it can sometimes help.

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: my concern with allowing lambda failures is the auto-retries. I suppose we should find the config param for that and terraform it so we can turn them off if desired.

I am starting to worry myself that it will actually be the gateway that throttles us and not the file shares.... in which case we will really have to think about it, possibly requiring the SNS model that we ditched a while back.

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: resolved via https://github.com/spacetelescope/calcloud/pull/114/commits/818b63e5eb83a356989659e1660a9a8fc95ceeff

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: I dug into the logs on Dev and I think that we are not being throttled by the gateway, but a run in test with a decent number of datasets is required for it to be definitive.

PR #121, opened 2021-05-28
batch-model-training

alphasentaurii:
semi-automated data ingest and compute resource prediction model training (memory bin classifier, memory regressor and wallclock regressor) via AWS batch. Data ingest and model re-training for large HST reprocessing runs can be deployed in one step by submitting a job in Batch. Docker image for batch compute env, job definition, and job queue are deployed via terraform.


modeling/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM library/ubuntu:20.04
+#SSL/TLS cert setup for STScI AWS firewalling
+USER root
+ENV REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+ENV CURL_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+RUN mkdir -p /etc/ssl/certs/ && mkdir -p /etc/pki/ca-trust/extracted/pem/
+COPY certs/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+RUN ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt && ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-certificates.crt && mkdir -p /etc/pki/ca-trust/extracted/openssl
+RUN apt update && apt install -y python3-dev python3-pip curl unzip && update-alternatives --install /usr/bin/python python /usr/bin/python3 10 && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && unzip awscliv2.zip && ./aws/install
```
bhayden53, line 9: I suspect this is why ECR is able to complain about python in your image, whereas the CALDP image has python in conda.

Not suggesting any changes right now, just thinking out loud on the ECR scan results.

terraform/model_training.tf
```diff
@@ -0,0 +1,70 @@
...
+
+  compute_resources {
+    allocation_strategy = "BEST_FIT"
+    instance_role = data.aws_ssm_parameter.ecs_instance_role.value
+    type = "EC2"
+    bid_percentage = 0
+    tags = {}
+    subnets             = local.batch_subnet_ids
+    security_group_ids  = local.batch_sgs
+    instance_type = ["c5a.2xlarge"]
```
bhayden53, line 33: curious what went into this decision. Do you use all the cores? Or do you need the memory?

jaytmiller (approved):
First time looking at this stuff but LGTM.

modeling/requirements.txt
```diff
@@ -0,0 +1,4 @@
+tensorflow==2.5.0
+sklearn==0.0
```
bhayden53, line 2: this seems odd. I googled it and came across this:
https://stackoverflow.com/a/40167445/1502840

Should there be an actual version number there and ubuntu gave something faulty at some point?

bhayden53: ah, I guess that really is the version of sklearn. Silly. 

I was just trying to understand if setting the version to 0.0 was some kind of special flag to pip.

terraform/model_training.tf
```diff
@@ -0,0 +1,71 @@
...
+    min_vcpus = 0
+    desired_vcpus = 0
+  }
+  lifecycle {
+    ignore_changes = [compute_resources.0.desired_vcpus]
+    create_before_destroy = true
+  }
+}
+
+resource "aws_batch_job_definition" "model_job_def_main" {
```
bhayden53, line 44: how does the job get triggered? Should there be a cron-like cloudwatch event (i.e. like the blackboard lambda)?

alphasentaurii: yes - we can add this later as a separate feature but for now we would manually submit as a batch job when needed

PR #123, opened 2021-06-01
auto rescues of DockerTimeoutError

bhayden53:
* change env var from "MAX_INSPECT_RETRIES" to "MAX_DOCKER_RETRIES" to be more generic
* auto-rescue "DockerTimeoutError", which is even more rare than the inspect errors

### DIFF
--- a/lambda/batch_events/batch_event_handler.py
+++ b/lambda/batch_events/batch_event_handler.py
@@ -70,12 +70,19 @@ def lambda_handler(event, context):
         else:
             print("Automatic OutOfMemory retries for", ipppssoot, "exhausted at", metadata["memory_retries"])
     elif container_reason.startswith("CannotInspectContainer"):
-        if not metadata["terminated"] and metadata["retries"] < int(os.environ["MAX_INSPECT_RETRIES"]):
+        if not metadata["terminated"] and metadata["retries"] < int(os.environ["MAX_DOCKER_RETRIES"]):
             metadata["retries"] += 1
             continuation_msg = "rescue-" + ipppssoot
             print("Automatic CannotInspectContainer rescue for", ipppssoot, "with retry count", metadata["retries"])
         else:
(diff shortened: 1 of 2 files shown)


lambda/batch_events/batch_event_handler.py
```diff
@@ -70,12 +70,19 @@ def lambda_handler(event, context):
...
             print("Automatic OutOfMemory retries for", ipppssoot, "exhausted at", metadata["memory_retries"])
     elif container_reason.startswith("CannotInspectContainer"):
-        if not metadata["terminated"] and metadata["retries"] < int(os.environ["MAX_INSPECT_RETRIES"]):
+        if not metadata["terminated"] and metadata["retries"] < int(os.environ["MAX_DOCKER_RETRIES"]):
             metadata["retries"] += 1
             continuation_msg = "rescue-" + ipppssoot
             print("Automatic CannotInspectContainer rescue for", ipppssoot, "with retry count", metadata["retries"])
         else:
             print("Automatic CannotInspectContainer retries for", ipppssoot, "exhausted at", metadata["retries"])
+    elif container_reason.startswith("DockerTimeoutError"):
```
jaytmiller, line 79: startswith() will actually take a tuple of strings.   so potentially you could collapse these two similar blocks of code using startswith and replacing the hard coded print strings with f"{container_reason}".  something to consider if they're basically the same handling.

jaytmiller (approved):
One possible simplification but otherwise LGTM

lambda/batch_events/batch_event_handler.py
bhayden53: I guess the only current difference is in the message that's logged; I could either make the message generic, or find a way to enumerate what's printed to match which startswith() was matched.

bhayden53:
If we add any more generic auto-rescues, I do think we should try to combine the startswith() calls into one clause. But we probably want to keep the error-specific logging messages, so we'd have to come up with a way to enumerate the logged messages.

PR #124, opened 2021-06-02
Add clean-ingested

jaytmiller:
Added a clean-ingested mode which restricts the datasets cleaned to those which have an ingested message;  in other respects it is similar to clean-all.

Potentially this could simplify / asbstract a little more by collapsing the all and ingested handlers,  but the "find ids" function varies so it's not quite as easy as it first appears.


bhayden53 (approved):
LGTM

PR #125, opened 2021-06-02
every ssm param explicitly nonsensitive (supporting upgrade to terraform 0.15.4)

bhayden53:
* ssm params seem to default to being sensitive. This causes terraform to complain about some outputs that are either directly sensitive, or contain a different sensitive value in their value. This can cause unpredictable behavior in some circumstances, and definitely can cause confusion about the reason that an output is being listed as sensitive.
* The cleanest way to deal with this is simply to mark every reference to an ssm_param as nonsensitive
* This is safe because:

1. There are no truly sensitive params (no keys or secrets)
2. the terraform output is not persisted anywhere, it is run in the CI node which is immediately torn down.


jaytmiller (approved):
What a PITA!   Could not find any other way either.  LGTM

PR #127, opened 2021-06-03
env-specific cache-refresh

bhayden53:
* overrides the long refresh time on inputs and control in dev and test, primarily for developers and testers who need quick turnaround on clean-all commands in order to start data processing runs on the same datasets

### DIFF
--- a/lambda/refreshCacheSubmit/refresh_cache_submit.py
+++ b/lambda/refreshCacheSubmit/refresh_cache_submit.py
@@ -38,7 +38,7 @@ def lambda_handler(event, context):
             print(str(exc))
 
     # may run twice but it's better than missing an entire hour
-    if do_infrequent_refresh(dt, os.environ['CALCLOUD_ENVIRONMENT']):
+    if do_infrequent_refresh(dt, os.environ["CALCLOUD_ENVIRONMENT"]):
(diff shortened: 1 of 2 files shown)


lambda/refreshCacheSubmit/refresh_cache_submit.py
```diff
@@ -47,3 +47,16 @@ def lambda_handler(event, context):
...
+
+
+def do_infrequent_refresh(timestamp, env):
+    # dev and test may require quick turnaround on runs and so are refreshed each time
+    override_envs = ["-dev", "-test"]
+
+    time_test = str(timestamp.minute)[0] in ("3")
+    env_test = any([s in env for s in override_envs])
+
+    if time_test or env_test:
```
jaytmiller, line 59: Never used any(),  that's really clean!    Here you can just do "return time_test or env_test" and save a little code.

jaytmiller (approved):
One possible simplification but otherwise LGTM

lambda/refreshCacheSubmit/refresh_cache_submit.py
bhayden53: good call, I will make that simplification

bhayden53:
oops, messed up the deploy script on the latest push

PR #130, opened 2021-06-16
refactor to remove docker builds from terraform

bhayden53:
* removes any docker builds from terraform and puts them in the deploy script; greatly speeds up the actual terraform apply and is generally more convenient for troubleshooting docker builds
* adds variables to deploy.sh so that local source dirs can be used for build (developer convenience)
* modifies blackboard lambda to fire every 7 minutes instead of 5 (hopefully dealing with snapshot file corruption)
* refactors model training Dockerfile for readability (confirmed the image is the same size)
* removed a step from the python build-from-source in the Dockerfile that took 5 minutes to run python tests
* bugfix for ingest.py when no data is scraped

### DIFF
--- a/modeling/Dockerfile
+++ b/modeling/Dockerfile
@@ -5,8 +5,37 @@ ENV REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
 ENV CURL_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
 RUN mkdir -p /etc/ssl/certs/ && mkdir -p /etc/pki/ca-trust/extracted/pem/
 COPY certs/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
-RUN ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt && ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-certificates.crt && mkdir -p /etc/pki/ca-trust/extracted/openssl
-RUN DEBIAN_FRONTEND=noninteractive && NCORES=`nproc` && apt update && apt upgrade -y && apt install -y build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev libsqlite3-dev wget libbz2-dev liblzma-dev curl unzip && wget https://www.python.org/ftp/python/3.9.5/Python-3.9.5.tgz && tar -xf Python-3.9.5.tgz && cd Python-3.9.5 && ./configure --enable-optimizations && make -j $NCORES && make altinstall && update-alternatives --install /usr/local/bin/python python /usr/local/bin/python3.9 10 && cd ../ && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && unzip awscliv2.zip && ./aws/install
+RUN ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt \
+&& ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-certificates.crt \
+&& mkdir -p /etc/pki/ca-trust/extracted/openssl
+
+RUN DEBIAN_FRONTEND=noninteractive && NCORES=`nproc` && apt update && apt upgrade -y && apt install -y \
+build-essential \
+curl \
+libbz2-dev \
+libffi-dev \
+libgdbm-dev \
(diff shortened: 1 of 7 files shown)


terraform/deploy.sh
```diff
@@ -18,37 +26,47 @@ CSYS_VER=`echo $CSYS_VER | awk '{print tolower($0)}'`
...
-tar -xvzf "v$CALCLOUD_VER.tar.gz"
-rm "v$CALCLOUD_VER.tar.gz"
-
-# caldp source download/unpack
-# github's tarballs don't work with pip install, so we have to clone and checkout the tag
-git clone https://github.com/spacetelescope/caldp.git
-cd caldp && git fetch --all --tags && git checkout tags/v${CALDP_VER} && cd ..
+# setting up the calcloud source dir if it needs downloaded
+# equivalent to "if len($var) == 0"
+if [ -z "$CALCLOUD_BUILD_DIR" ]
```
alphasentaurii, line 34: Not a huge deal but maybe would put the mkdir and cd for $TMP_INSTALL_DIR inside the conditional (if I set $CALCLOUD_BUILD_DIR to something the script still builds and changes into a tmp dir before checking if a custom build dir was set) i.e. these come before the conditional but should go inside it
`# variables that will be changed less-frequently
TMP_INSTALL_DIR="/tmp/calcloud_install"
mkdir $TMP_INSTALL_DIR
cd $TMP_INSTALL_DIR`

modeling/Dockerfile
```diff
@@ -5,6 +5,37 @@ ENV REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
...
+&& cd Python-3.9.5 \
+&& ./configure --enable-optimizations \
+&& make altinstall \
+&& update-alternatives --install /usr/local/bin/python python /usr/local/bin/python3.9 10 
+
+RUN DEBIAN_FRONTEND=noninteractive \
+&& cd ../ && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
+&& unzip awscliv2.zip \
+&& ./aws/install
 RUN ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt && ln -s /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-certificates.crt && mkdir -p /etc/pki/ca-trust/extracted/openssl
```
alphasentaurii, line 39: You've got the certs in there twice it looks like: lines8-10 and 39 (causes an error during build bc file already exists)

modeling/Dockerfile
```diff
@@ -5,6 +5,37 @@ ENV REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
...
 RUN DEBIAN_FRONTEND=noninteractive && NCORES=`nproc` && apt update && apt upgrade -y && apt install -y build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev libsqlite3-dev wget libbz2-dev liblzma-dev curl unzip && curl -o Python-3.9.5.tgz https://www.python.org/ftp/python/3.9.5/Python-3.9.5.tgz && tar -xf Python-3.9.5.tgz && cd Python-3.9.5 && ./configure --enable-optimizations && make -j $NCORES && make altinstall && update-alternatives --install /usr/local/bin/python python /usr/local/bin/python3.9 10 && cd ../ && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && unzip awscliv2.zip && ./aws/install
```
alphasentaurii, line 40: Need to delete line 40 as well (duplication causes everything to install twice) - BTW you can get rid of the NCORES=`nproc` arg since we're no longer running the command that calls it ("make -j $NCORES")

alphasentaurii (changes requested):
couple of duplicate lines in Modeling docker file that should be removed. Potential minor revision to deploy.sh (non-critical)

terraform/deploy.sh
bhayden53: fixed via https://github.com/spacetelescope/calcloud/pull/130/commits/b6b7380b3fd47db8e2dfe43373bf7535e1ce41d9

I ended up just putting a mkdir in each conditional with a -p, although probably it's more straightforward to do an or on the two custom directories to see if the tmp directory is needed at all.

modeling/Dockerfile
bhayden53: also addressed in https://github.com/spacetelescope/calcloud/pull/130/commits/67683ff8ab5d03cd632ff8996deda972b87bb3ed

modeling/Dockerfile
bhayden53: good catch, I think this was due to the merge issue I had with vs-code. fixed in https://github.com/spacetelescope/calcloud/pull/130/commits/67683ff8ab5d03cd632ff8996deda972b87bb3ed

alphasentaurii (approved):

PR #132, opened 2021-06-23
restrict S3 invocation of lambda's to proper source account owner

bhayden53:
* permissions hardening of S3 event-based lambda invocation


jaytmiller (approved):
Looks clean.  Seems complete.

PR #134, opened 2021-07-01
lambda-data-ingest

alphasentaurii:
-s3 message ('processed-{ipst}.trigger') triggers ingest lambda (lambda/ingest/lambda_scrape.py)
-ingest lambda iteratively scrapes job data (features and targets) from s3 bucket (/control, outputs/{logfiles}) then uploads to DynamoDB table
-batch modeling (docker image) scans and downloads data from DB, trains models, then uploads model zipfiles and results metadata to s3 bucket (calcloud-modeling)


bhayden53:
first thoughts: we'll need to be more specific about what "ingest" means as that has very strong prior meaning for archive folks. Also we should be able to terraform the creation of the zip package; I can sort that out.

go ahead and merge and I'll pick it up and start working on those things.

bhayden53 (approved):

PR #135, opened 2021-07-02
initial refactor of new model ingest lambda

bhayden53:
* terraforms the zip package creation by adding a requirements.txt to the lambda source directory
* renames lambda and ddb
* NOTE: do we want a new IAM role for this lambda?
* I had to make a couple of changes to the deploy role to get this version to deploy.


bhayden53:
if possible, we should pin the version of sklearn in requirements

alphasentaurii (approved):

PR #139, opened 2021-07-20
ddb-batch-model-retraining

alphasentaurii:
-modularized code into separate scripts for prep, training, validation, etc
-added ability to scan ddb table using filter expression
-save model metadata as pickle objects (for easier export into dashboard)
-fix intermittent kfold crossval memory error (n_jobs can be set as env var depending on size of dataset)
-new IAM role for model ingest
-renamed JobPredict docker image tag "predict" instead of "model"
-changed DDB billing type from provisioned to On-demand


bhayden53:
lgtm

bhayden53 (approved):

PR #145, opened 2021-08-02
Dynamodb batch

alphasentaurii:
Memory and Wallclock prediction models retrieve training data from DynamoDB and are retrained via Batch model training job. Learning metadata continues to be stored in s3.


lambda/ModelIngest/lambda_scrape.py
```diff
@@ -0,0 +1,334 @@
+import boto3
```
bhayden53, line 1: I wonder if it makes sense to put some of these functions in calcloud and just keep the lambda_handler in here

modeling/io.py
```diff
@@ -2,47 +2,148 @@
...
 import json
+import csv
 import numpy as np
 import zipfile
+from boto3.dynamodb.conditions import Attr
+import pickle
 
 # mitigation of potential API rate restrictions (esp for Batch API)
-retry_config = Config(retries={"max_attempts": 5, "mode": "standard"})
+retry_config = Config(retries={"max_attempts": 5})
```
bhayden53, line 13: now that I've seen this twice, I wanted to point out that we have one of these in calcloud.common; maybe if you want a retry config with less max_attempts, you could add another entry in common. Or maybe even we could make the common entry a function that takes max attempts as a variable.

The key value there being if we ever decide there's something else we want to add to every config item, we only need to change it in that one place.

lambda/ModelIngest/lambda_scrape.py
alphasentaurii: the lambda has access to what's in calcloud?

bhayden53: it does when you do this in the terraform:
https://github.com/spacetelescope/calcloud/blob/67344b895f757bab780a8805d7da7ea717c54a5f/terraform/lambda_job_submit.tf#L24

bhayden53: I guess I am a proponent of keeping the lambda script itself as lightweight as possible. Basically just that lambda_handler, iff possible. It helps with readability in the AWS console.

PR #146, opened 2021-08-02
tensorflow2.5

alphasentaurii:
Updated JobPredict lambda to use Tensorflow 2.5.0; re-saved same version of models that are currently in Ops with Tensorflow 2.5.0 (no change to model weights/variables - simply adds additional file for Keras metadata).


bhayden53 (approved):
lgtm

PR #147, opened 2021-08-04
s3-backup-training-data

alphasentaurii:
After training data is retrieved from DynamoDB, it's saved into a local file named "latest.csv" and uploaded to a persistent s3 bucket (calcloud-modeling) as a backup.


bhayden53 (approved):
lgtm

PR #148, opened 2021-08-05
refactor/model-ingest

alphasentaurii:
Moved functions to separate script inside main calcloud directory "model_ingest.py"; ModelIngest/lambda_scrape.py is much cleaner with only the lambda_handler function which initiates dynamodb ingest in a single command.


bhayden53 (approved):
lgtm

PR #149, opened 2021-08-09
Hotfix/v0.4.28

bhayden53:
- the AWS API call to Batch for "container" "reason" can apparently include newline characters. We were dumping this string straight to the blackboard snapshot file, and the newline was breaking the on-premise poller. This hotfix replaces newline characters in the "exitReason" string with an empty string.


jaytmiller (approved):
LGTM

PR #151, opened 2021-08-23
Model updates

alphasentaurii:
- Models retrained on data through 8-20-2021; model files updated in-place. 
- PowerTransform values updated (relates to distribution of n_files and total_mb features)
- Minor fix for crsplit value in JobPredict: allows for floating point (1.0) or integer (1)


bhayden53 (approved):
lgtm

PR #158, opened 2021-09-15
fix/model-ingest-lambda-config

alphasentaurii:
Increased memory size to 256mb for model-ingest lambda (usage hits 125-128mb on avg)
Print ipst name to log in case of errors during s3 input scraping (for easier troubleshooting)
Added scripts for manually moving s3 backup csv files to and from DDB envs


bhayden53 (approved):
lgtm

PR #159, opened 2021-10-05
Add message payloads,  job timeout_scale factor

jaytmiller:
Added support for YAML payloads in messages,  extended broadcasting to  
  handle a single common payload for broadcasts (rescue-all).   At the method API level
  the payload is just a serializable Python object being stored or loaded.
  As message text,  the payload object is serialized in YAML.
Modified submit and rescue lambdas to use the payload from the trigger
  message to define job overrides.
Added timeout_factor override to submit code and planner which
  scales the modeled job time by the given floating point value.   Note
  that Batch has a min kill time of 60 seconds so some jobs cannot
  be scaled down far enough to fail.
Refactored broadcast lambda and io to improve testability.
Added pyyaml pip package to calcloud.io-dependent lambdas.   This included
  adding a requirements.txt file in the calcloud package directory,  as
  well as Terraform mods to refer to it.
Updated Python environment in calcloud.io-dependent lambdas to 3.7 to
  match the system python installed on CI-nodes; this is necessary
  because the system python is used to compile pip package byte codes
  for inclusion in the lambda zip file.
Turned off log timestamps since CloudWatch adds them anyway;  kept changes
  minimal although using logging vs. print to flag INFO, ERROR, or WARNING
  states might be useful.


calcloud/plan.py
```diff
@@ -108,18 +104,22 @@ def _get_resources(ipppssoot, output_bucket, input_path):
     job_name = ipppssoot
     input_path = input_path
     crds_config = "caldp-config-aws"
-    # invoke calcloud-ai lambda
-    predictions = invoke_lambda_predict(ipppssoot, output_bucket)
-    initial_bin = predictions["memBin"]  # 0
-    kill_time = min(max(predictions["clockTime"] * 5, 20 * 60), 48 * 60 * 60)  # between 20 minutes and 2 days
+
+    clockTime, initial_bin = invoke_lambda_predict(ipppssoot, output_bucket)
+    # clockTime, initial_bin = 600, 0
```
bhayden53, line 109: dev leftover?

setup.cfg
```diff
@@ -13,12 +13,13 @@ github_project = spacetelescope/calcloud
 
 [options]
 packages = find:
-python_requires = >=3.8
+python_requires = ~=3.7.0
 setup_requires =
     setuptools >=41.0.1
 install_requires =
     boxsdk[jwt] >=2.5.0, <3
```
bhayden53, line 20: why is the boxsdk a requirement? I think that might be a really, really old bug?

terraform/lambda_model_ingest.tf
```diff
@@ -39,7 +39,12 @@ module "lambda_model_ingest" {
       path = "${path.module}/../calcloud"
       prefix_in_zip = "calcloud"
       pip_requirements = false
-    }
+    },
+    #  {
```
bhayden53, line 43: any need to leave the commented out bit?

setup.cfg
jaytmiller: Yeah...   I think maybe setup.cfg and tox were cloned form the URI redirector.

terraform/lambda_model_ingest.tf
jaytmiller: Nothing iron clad.  My intent was to leave it as a warning/clue in case the lambda ever does start using the io stuff or anything which triggers the import of io.

jaytmiller: Also as a reminder that adding it didn't work.

bhayden53 (approved):
LGTM, thanks!

terraform/lambda_model_ingest.tf
bhayden53: I'm good with leaving it in for those reasons.

PR #161, opened 2021-10-06
query ddb for wallclock and std_err multiplier

alphasentaurii:
Queries dynamo_db table for pre-existing record of wallclock time and pre-calculated standard error. If record does not exist, defaults to 20 minutes and 5, respectively. The wallclock prediction is multiplied by 1 + standard error; whichever one of these is the maximum (20 min vs prediction*1+err) will be the kill time used by the job.


bhayden53 (approved):
lgtm

PR #164, opened 2021-10-29
model-ddb-updates

alphasentaurii:
-Models updated in place (trained on 240k dataset from Ops)
-model training saves pt_transform data to json file inside of models.zip
-model training updates DDB with predictions, x_files, x_size, and wallclock subsample stats (mean, std, std_err)
-JobPredict lambda loads pt_transform data from json file
-modelIngest no longer needs pt_transform data


calcloud/plan.py
```diff
@@ -78,13 +78,13 @@ def query_ddb(ipppssoot):
...
     response = table.query(KeyConditionExpression=Key("ipst").eq(ipppssoot))
-    db_clock, std_err = 20 * 60, 5
+    db_clock, wc_std = 20 * 60, 5
     if len(response["Items"]) > 0:
         data = response["Items"][0]
-        db_clock = int(data["wallclock"])
-        if "std_err" in data:
-            std_err = float(data["std_err"])
-    return db_clock, std_err
+        db_clock = float(data["wallclock"])
```
bhayden53, line 84: Did you test some actual batch jobs with this change? I just have a minor concern that it won't accept a float, and I'm not sure if this is changed to an int somewhere else

lambda/JobPredict/predict_handler.py
```diff
@@ -126,11 +133,14 @@ def transformer(self):
         # apply power transformer normalization to continuous vars
         x = np.array([[n_files], [total_mb]]).reshape(1, -1)
         pt = PowerTransformer(standardize=False)
-        pt.lambdas_ = np.array([-1.05989146, 0.1691683])
+        pt.lambdas_ = np.array([pt_data["f_lambda"], pt_data["s_lambda"][1]])
```
bhayden53, line 136: pt_data["s_lambda"][1]

Is that index a bug?

calcloud/plan.py
alphasentaurii: yes it's been tested, and yes it gets converted to an int in Plan.py

lambda/JobPredict/predict_handler.py
alphasentaurii: fixed

PR #166, opened 2021-11-01
Override and metadata parameter interchangeability

jaytmiller:
Refactor submit message payload handling to make stored metadata vs. override
  parameters more interchangeable.
Add memory_bin parameter as absolute override of bin,  used instead of model or
  memory_retries increments.
Pass all metadata parameters into planner to reduce changes required as new
  planning parameters are added in the future.


bhayden53 (approved):
LGTM

PR #167, opened 2021-11-12
Add batch event tests

jaytmiller:


tests/test_io.py
```diff
@@ -0,0 +1,12 @@
+from . import conftest
+
+_fakeout = conftest
```
bhayden53, line 3: What's this?

bhayden53 (approved):
Looks great

tests/test_io.py
jaytmiller: That's to avoid failing code quality tests due to not explicitly using conftest anywhere else.   I'm sure there's a better way now that you point it out.

PR #171, opened 2021-12-02
[WIP] first refactor of docker builds into separate script

bhayden53:
in prep for central ecr refactor.

This refactor does not change the calling sequence for deployment yet, it simply factors some variable setting into a script, and factors the docker builds/pushes into a script. This will make it much simpler to implement the central ecr in a future refactor.

### DIFF
--- a/terraform/deploy_docker_builds.sh
+++ b/terraform/deploy_docker_builds.sh
@@ -10,9 +10,9 @@ source deploy_checkout_repos.sh
 # and it will be set in the deploy_vars.sh script
 # if you want to run this script independently of deploy.sh, you'll need to set repo_url in the env manually
 ##### DOCKER IMAGE BUILDING #########
-CALDP_DOCKER_IMAGE="${repo_url}:batch-${COMMON_IMAGE_TAG}_dev"
-PREDICT_DOCKER_IMAGE="${repo_url}:unscanned-predict-${COMMON_IMAGE_TAG}_dev"
-TRAINING_DOCKER_IMAGE="${repo_url}:unscanned-training-${COMMON_IMAGE_TAG}_dev"
+CALDP_DOCKER_IMAGE="${repo_url}:batch-${COMMON_IMAGE_TAG}-dev"
+PREDICT_DOCKER_IMAGE="${repo_url}:unscanned-predict-${COMMON_IMAGE_TAG}-dev"
+TRAINING_DOCKER_IMAGE="${repo_url}:unscanned-training-${COMMON_IMAGE_TAG}-dev"
 
 # naming is confusing here but "modeling" directory plus "training" image is correct
 cd ${CALCLOUD_BUILD_DIR}/modeling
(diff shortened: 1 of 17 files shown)


bhayden53:
@jaytmiller any thoughts?

terraform/deploy_docker_builds.sh
```diff
@@ -0,0 +1,44 @@
+source deploy_vars.sh
+
+# this is the tag that the image will have in AWS ECR
```
jaytmiller, line 3: There  may be some value in tagging it with both latest and the release version.   Latest will go away when the next image steals it,  but the release version would remain identifying the otherwise anonymous image hash.

terraform/deploy_vars.sh
```diff
@@ -0,0 +1,29 @@
+export CALCLOUD_VER="v0.4.31"
```
jaytmiller, line 1: Factoring this out like this is what I'd do too.

jaytmiller (approved):
LGTM

terraform/deploy_docker_builds.sh
bhayden53: I'm planning to totally refactor the tagging during the next step

bhayden53:
@jaytmiller I've significantly refactored since you approved. Would you mind taking another look?

terraform/deploy_existing_imageTag_check.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo "usage: deploy_check_existing_imageTag.sh <tag1 tag2 tag3 ...>"
+    echo
+    exit 2
+fi
+
+anyImage="0"
+for imageTag in ${TAGS}; do
+    # describe images returns exit code 1 if the image does not exist
+    # batch-get-image does NOT
+    awsudo ${ADMIN_ARN} aws ecr describe-images --registry-id ${ECR_ACCOUNT_ID} --repository-name ${IMAGE_REPO} --image-ids imageTag=${imageTag}
```
jaytmiller, line 20: For future reference,  I think you can write this more like:

try-to-list-tag  && echo "tag exists" && anyImage="1"

terraform/locals.tf
```diff
@@ -80,9 +80,21 @@ locals {
               },
        ]
 
-       ecr_address = format("%v.dkr.ecr.%v.amazonaws.com", data.aws_caller_identity.this.account_id, data.aws_region.current.name)
-       ecr_predict_lambda_image   = format("%v/%v:predict", local.ecr_address, aws_ecr_repository.caldp_ecr.name)
-       ecr_model_training_image = format("%v/%v:training", local.ecr_address, aws_ecr_repository.caldp_ecr.name)
+       # should be kept in sync with the string in deploy_vars.sh
```
jaytmiller, line 83: Nice locals!

jaytmiller (approved):
Wow,  that was a monster.   LGTM!

PR #175, opened 2022-02-14
modeling/jan-2022-repro

alphasentaurii:
updated model files in place from Jan2022 repro retraining. Results and dataset available on s3: 2022-02-14-1644848448


calcloud/io.py
```diff
@@ -43,7 +43,7 @@
     "clean",
 ]
 
-MAX_BROADCAST_MSGS = 10 ** 6  # safety
+MAX_BROADCAST_MSGS = 10**6  # safety
```
jaytmiller, line 46: Did black make these whitespace changes?

jaytmiller (approved):
Careful with Black but otherwise LGTM.

calcloud/io.py
jaytmiller: nvm,  missed that commit.  black == love/hate,  mostly hate!

alphasentaurii: Yes it must have - Not sure how Black failed the GH actions check on my first commit since I didn't touch any code and was using the latest version of develop branch...anyhow I upgraded to a newer version of black and re-ran, resulting in this revision - other than that only changes I made for this PR was replacing saved model files.

bhayden53: this happens sometimes when black is updated and new versions start getting pulled into the github action run. We could hard-pin black perhaps.

bhayden53:
LGTM but before deleting the source branch could you PR to the open release branch as well? I'm not sure what else could be on develop but I want to get this into the next build.

bhayden53:
don't merge this, I will get it to dev when the build is released to main, and merging this to dev now will cause a merge conflict for no good reason

PR #185, opened 2022-07-22
WIP additional tests

bhayden53:
* ami_rotation lambda (just runs an ec2)
* clean lambda
* delete lambda
* rescue lambda
* hst.py


tests/test_hst.py
```diff
@@ -0,0 +1,27 @@
+from calcloud import hst
+from . import conftest
+
+IPPPSSOOT_INSTR = hst.IPPPSSOOT_INSTR
+INSTRUMENTS = set(IPPPSSOOT_INSTR.values())
```
bhayden53, line 5: @jshihstsci I would grab this one from hst.py as well. If someone changed the code in hst.py for constructing this object, you'd have to change the code in the test to match.

PR #198, opened 2023-06-13
Release/v0.4.40

raswaters:


terraform/variables.tf
alphasentaurii, line 1: Current ops context is hst_1078.pmap (as of 6/9) - not sure if this needs to be updated or if the default setting is just a fallback?

bhayden53: default setting is a fallback; the crds sync updates a file in the storage gateway to indicate the actual current context to the cloud. 

We still have time to update this if we want to though. 

It's good for the install plan to have a step to check the context for the cloud though.

bhayden53 (approved):

bhayden53:
We need to wait for a new base image; matplotlib needs downgraded for spectral previews in caldp

terraform/variables.tf
raswaters: Because the default is a fall-back and because I'll update this in the next release I'm about to start, I'm resolving this conversation.

PR #203, opened 2024-01-05
Update ssl cert workflow

jshihstsci:


jaytmiller (approved):
This looks good to me but I should point out a few things:
1. The need to check and clean the cert dates to the first iteration of the new S3 cert from ITSD when they initially messed up the format.  Not knowing what it really should be or what they intended, I just fixed it.  Later they fixed it.   So the listing/fixing/checking functionality is not really required.  Hence I believe there is the option of simplifying and just using copy-cert from codebuild,  although personally I find cert-list useful when encountering new certs.
2. This may be obvious somewhere but make sure the cert is not included in the git repo,  it is considered to be sensitive even though it is a "public" part of the crypto because it is only exposed on internal networks.
3. Ultimately "working" is the best test.  Make sure Tanner has packet inspection turned on.
4. While it all looks good to me and working/building == success,  Manuel did the codebuild "port" so he may have other valuable comments.

msanchezst (approved):
LGTM. As Todd mentions working/building == success 
Thanks!

jshihstsci:
Thanks for the reviews! This has been tested on dev and worked fine.

msanchezst:
> Thanks for the reviews! This has been tested on dev and worked fine.
With DPI enabled?

jshihstsci:
> Thanks for the reviews! This has been tested on dev and worked fine.
> With DPI enabled?

Not sure if DPI was enabled, is there a way I can check?

msanchezst:
> > Thanks for the reviews! This has been tested on dev and worked fine.
> > With DPI enabled?
> 
> Not sure if DPI was enabled, is there a way I can check?

I guess either check with the networking team or perhaps run a test without the cert included.

jshihstsci:
> > > Thanks for the reviews! This has been tested on dev and worked fine.
> > > With DPI enabled?
> > 
> > 
> > Not sure if DPI was enabled, is there a way I can check?
> 
> I guess either check with the networking team or perhaps run a test without the cert included.

I tried building the codebuild image and it failed to build without STSCICA.crt (no problem with building when the cert is included). Can we then assume that DPI is enabled and the cert is doing what it's supposed to?

jaytmiller:
I think so.  You should be able to verify  DPI with Tanner pretty quick.    STSCICA.crt probably needs to be updated from S3 on every build to pick up periodic changes from ITSD as well as just being available w/o inclusion in the repo.   I'd also turn off the exception trapping for the S3 update routines.   (That may pre-date ITSD full support for the S3 but in any case the EFS Quota "cert template" is allowed to include the cert because it is CodeCommit vs. public GitHub.  So it is not a good example with regard to cert updates.)

jshihstsci:
I updated cert-update so that it will fail if it cannot get the most recent cert from S3, and also removed the older certs that we used to include in the repo. Leaving clean_cert and cert-list in for now since they shouldn't hurt if certs are formatted properly. I'll verify the DPI status with Tanner and run more tests if necessary.

PR #209, opened 2024-12-05
update codebuild base to al2023

jshihstsci:


raswaters (approved):
Is that symbol below line 57 and indication that there is no newline at the end? Does that matter?

PR #210, opened 2025-02-11
Update tag-latest.yml

pllim:
As recommended by actions/runner-images#11101 , this PR replaces the deprecated ubuntu-20.04 runner with a newer version.

Also upgrades checkout and use hash pin for security.

[:ghost:](https://github.com/pllim/playpen/issues/63)


sashawood:
New tickets created to do this work:
https://jira.stsci.edu/browse/CALCLOUD-482
https://jira.stsci.edu/browse/CALCLOUD-483

PR #218, opened 2026-06-24
CALCLOUD-464 Make amirotate codebuild script work again

sashawood:
Do not pass the output of `aws ec2 describe-images` on the command line. It is too long for the OS to handle.  Instead, pass it through a file.

I noticed that the CodeBuild project calcloud-ami-rotation-ops was failing in all accounts (CodeBuild > Build > Build projetcs > calcloud-ami-roration-ops): https://us-east-1.console.aws.amazon.com/codesuite/codebuild/812556236571/projects/calcloud-ami-rotation-ops/history

I think there are two reasons this is failing.  The easier one is that in three files, we call `parse_image_json.py` with the output of `aws ec2 describe-images`.  This output is about 400K, and that is too long to pass on the command-line. 
- deploy.sh
- deploy_ami_rotate.sh - I think this is unused
- deploy_ami_rotate_codebuild_script.sh

@raswaters, you made this change to correct it in deploy.sh, but did not make the change in the other locations:
https://github.com/spacetelescope/calcloud/commit/fe1a6a7fecf9f2d856c7c3a7d0285c71e010d13c

I applied your fix to deploy_ami_rotate_codebuild_script.sh in the sandbox account, but I had problems testing because of bash quoting madness.  I changed the code to instead pass the json output through a file and verify that works.  I think this is a better solution, so I made the change in the deploy.sh script as well.

### DIFF
--- a/ami_rotation/parse_image_json.py
+++ b/ami_rotation/parse_image_json.py
@@ -3,9 +3,24 @@ import json
 from collections import OrderedDict
 from datetime import datetime
 
-response = json.loads(str(sys.argv[1]))
+# This command parses the response to `aws ec2 describe-images`.
+# This response can be so long that it cannot be passed on the command-line.
+# When you call this program, you can either:
+# 1. pass the response on the command line, followed by the image-name-prefix
+# 2. store the response in a file called images.json, and only pass the image-name-prefix
+if len(sys.argv) == 3:
+    # Response has been passed on command-line
+    input_string = str(sys.argv[1])
+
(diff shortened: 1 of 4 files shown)


raswaters:
I made the changes under fe1a6a7 to get past the immediate problem blocking the deploy. I am fine with the changes made here. Slight downside is that it pulls in a very long list of images, but downloading 0.5MB and parsing it is fast.

raswaters (approved):

jshihstsci (approved):

PR #219, opened 2026-07-17
CALCLOUD-462 Increase disk size for reprocessing MVMs

sashawood:
Previously, the terraform scripts:
* created 1 launch template called hstsdp with a disk size of 150GB.
* used a variable called local.ladder with 4 rungs - 02g, 08g, 16g, 64g
* created 4 job queues, compute environments, and job definitions - one each for each of the 4 local.ladder rungs

I have modified the scripts to:
* create 2 launch templates: hstdp["default"] with a disk size of 150GB and hstdp["large-volume"] with a disk size of 300GB.
* use a local.ladder var with 8 rungs - 02g, 08g, 16g, 64g and 02g-large-volume, 08g-large-volume, 16g-large-volume, 64g-large-volume
* create 8 job queues, compute environments, and job definitions - one each for each of the local.ladder rungs

In addition, 
* I modified multiple places where we explicitly used to iterate over 4 rungs of local.ladder to now use iterate over length(local.ladder) rungs.
* In plan.py, I modified functions to use "default" AWS resources for ipppssoots and svms, and "large-volume" resources for mvms.
* I updated the tests to include the new queues, compute environments, and definitions.

I have run this with tests 08106 and 08107 and observed the ipppssoots and svms are routed to default volume size EC2 instances, and mvms are routed to large-volume EC2 instances. 
https://jira.stsci.edu/browse/CALCLOUD-462?focusedId=1037304&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1037304

### DIFF
--- a/terraform/batch.tf
+++ b/terraform/batch.tf
@@ -35,19 +35,38 @@ terraform {
 
 # See also lambda module version in each lambda .tf file
 
+# We must use a different launch template for each volume size we use.
+locals {
+  launch_templates = {
+    default = {
+      name_suffix       = ""
+      volume_size       = 150
+      description_infix = ""
+    }
+    large-volume = {
+      name_suffix       = "-large-volume"
+      volume_size       = 3000
+      description_infix = "large-volume "
+    }
+  }
+}
+
 resource "aws_launch_template" "hstdp" {
   # IF YOU CHANGE THE LAUNCH TEMPLATE YOU MUST "TAINT" THE COMPUTE ENVIRONMENT BEFORE DEPLOY
   # IN ORDER FOR YOUR CHANGES TO BE PICKED UP BY BATCH
   # See https://docs.aws.amazon.com/batch/latest/userguide/create-compute-environment.html
   # and https://github.com/hashicorp/terraform-provider-aws/issues/15535
-  name = "calcloud-hst-worker${local.environment}"
-  description             = "Template for cluster worker nodes updated to limit stopped container lifespan"
+
+  for_each = local.launch_templates
+
+  name                   = "calcloud-hst-worker${each.value.name_suffix}${local.environment}"
+  description            = "Template for ${each.value.description_infix}cluster worker nodes updated to limit stopped container lifespan"
   ebs_optimized           = "false"
   image_id                = nonsensitive(aws_ssm_parameter.ecs_ami.value)
   update_default_version = true
   tags = {
-    "Name"            = "calcloud-hst-worker${local.environment}"
-    "calcloud-hst"    = "calcloud-hst-worker${local.environment}"
+    "Name"            = "calcloud-hst-worker${each.value.name_suffix}${local.environment}"
+    "calcloud-hst"    = "calcloud-hst-worker${each.value.name_suffix}${local.environment}"
     "stsci-poc-email" = var.stsci_poc_email
   }
   user_data               = base64encode(templatefile("${path.module}/user_data.sh", {}))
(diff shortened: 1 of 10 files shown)


terraform/batch.tf
```diff
@@ -125,7 +144,7 @@ resource "aws_batch_compute_environment" "compute_env" {
     desired_vcpus = local.ladder[count.index].ce_desired_vcpus
 
     launch_template {
-      launch_template_id = aws_launch_template.hstdp.id
+      launch_template_id = aws_launch_template.hstdp[local.ladder[count.index].lt_volume_size_key].id
```
raswaters, line 147: What does the lt_volume_size_key do? Given the absence of an index before, was this always returning the same template id regardless of where we were on the ladder?

terraform/deploy.sh
```diff
@@ -43,13 +43,18 @@ awsudo $ADMIN_ARN terraform init -backend-config="bucket=${aws_tfstate}" -backen
 cd ${CALCLOUD_BUILD_DIR}/terraform
 
 # must taint the compute env to be safe about launch template handling. see comments in batch.tf
-awsudo $ADMIN_ARN terraform taint aws_batch_compute_environment.compute_env[0]
-awsudo $ADMIN_ARN terraform taint aws_batch_compute_environment.compute_env[1]
-awsudo $ADMIN_ARN terraform taint aws_batch_compute_environment.compute_env[2]
-awsudo $ADMIN_ARN terraform taint aws_batch_compute_environment.compute_env[3]
+for i in {0..7}; do
```
raswaters: 0 to 7 instead of 0 to 3 because 4 large mvm ones were added?

terraform/batch.tf
raswaters: Never mind, I see the lt_volume_size_key below

tests/conftest.py
```diff
@@ -25,9 +25,19 @@
 
 EVENT_DIR = str(Path(__file__).resolve().parent / "artifacts/events")
 
-JOBDEFINITIONS = ["calcloud-jobdef-2g", "calcloud-jobdef-8g", "calcloud_jobdef-16g", "calcloud-jobdef-64g"]
-CENVIRONMENTS = ["calcloud-cenv-2g", "calcloud-cenv-8g", "calcloud-cenv-16g", "calcloud-cenv-64g"]
-JOBQUEUES = ["calcloud-jobqueue-2g", "calcloud-jobqueue-8g", "calcloud-jobqueue-16g", "calcloud-jobqueue-64g"]
+JOBDEFINITIONS = ["calcloud-jobdef-2g", "calcloud-jobdef-2g-large-volume",
+                  "calcloud-jobdef-8g", "calcloud-jobdef-8g-large-volume",
+                  "calcloud_jobdef-16g", "calcloud_jobdef-16g-large-volume",
+                  "calcloud-jobdef-64g", "calcloud-jobdef-64g-large-volume"
```
raswaters: Is it convenient for development/implementation to have four memory buckets for mvm? I am not sure how well the memory model has been trained on mvm data. There are mvms with only a few exposures, and for those a small amount of memory likely is enough. A key factor is also what the spread on the sky is of those exposures. Two exposures that are on opposite sides of a skycell will result in more memory and disk space being used. Just wondering, I think this is fine as is.

raswaters (approved):

tests/conftest.py
sashawood: I don't understand the requirements or behavior well enough to have an opinion yet.

terraform/deploy.sh
sashawood: yes, that's correct.

PR #222, opened 2026-07-24
CALCLOUD-474 Add script to calculate models for memory and runtime

sashawood:
I have developed some simpler models that get reasonable results for memory and wallclock for our data.

I used the first 100,000 records from the DynamoDB table calcloud-model-ops as training and testing data.  This a random sampling of all the data we have.

I found that a rainforest regression gives very good results for memory.  Rainforest regressions are available in scikit-learn - we don't have to use tensorflow for them.
```
Memory
  R²              = 0.981
  MAE          = 0.04
  RMSE        = 0.09
  Mean APE = 5.0%
```

For the wallclock, I used a linear regression to determine log(wallclock), and I used two interaction terms - one for total_mb * pctecorr and one for n_files * pctecorr.

The model is pretty good, but overestimates wallclock time on some large (> 250MB) files.
```
Wallclock
  Log-space R²   = 0.943
  Log-space MAE  = 0.26
  Log-space RMSE = 0.39
  Mean APE       = 29.6%
  Median APE     = 17.3%
  95th pct APE   = 88.3%
  99th pct APE   = 217.6%
```
I tried a rainforest regression, I tried using log(total_mb) and log(n_files), and none of these were improvements.

What do you think about using these models to replace the machine learning models we're using?

* I would use the memory prediction + 10% to bin into machines with 2GB, 8GB, 16GB, and 64GB memory.
* A comment in the current code says that it takes the wallclock prediction and multiplies x 3 to get the timeout for AWS batch.  I haven't found the place in the code that actually does that, but I would verify that is what is happening, and do a same with the new predictions.


raswaters:
As we discussed earlier this week, I am very much in favor of using a linear regression because I think that will simplify our code base, dependencies, and unit tests. The correlations you found appear strong. Would be nice to see some visualizations. I like the idea of adding some memory to the value found solely based on the regression to reduce the need for escalation up the memory ladder.

Rainforest = random forest?

sashawood:
I looked at it with fresh eyes today, and I have a better fit for wallclock.

Memory uses a RandomForestRegressor (not rain forest, as you noticed).  These are the same numberes as above
  R²              = 0.981
  MAE          = 0.04
  RMSE        = 0.09
  Mean APE = 5.0%

Wallclock now uses a HistGradientBoostingRegressor, which is a better model for the data.
  Log-space R²   = 0.981
  Log-space MAE  = 0.14
  Log-space RMSE = 0.22
  Mean APE       = 14.1%
  Median APE     = 7.3%
  95th pct APE   = 47.8%
  99th pct APE   = 91.7%
  
Distribution-all
I printed out a histogram for wallclock and memory.  Both are exponential in nature.  
* 50% of the wallclock values < 60s, which is the minimum that AWS BatchJob lets us set for a timeout. 
* more than 90% of the memory values < 2G, which is the smallest memory we can use.  


Here are graphs for wallclock and memory actual vs. predicted.  If the prediction were perfect, all the points would be on the red line.

Wallclock - using a log scale because results span 10 - 10000 s
Wallclock-predicted-v-actual

Memory
Memory-predicted-v-actual

sashawood:
@jshihstsci wondered how the fit looks for the existing ML model, so I pulled some data.

Here's the same stats as for memory above.  ML is worse than the model I suggest.
  R²   = 0.638
  MAE  = 0.25
  RMSE = 0.39
  MAPE = 25.3%

And you can see the difference in predicted vs. actual:
memory-by-ml


Here's the same stats for wallclock above.  The models are equivalent.
  Log-space R²   = 0.970
  Log-space MAE  = 0.20
  Log-space RMSE = 0.00
  Mean APE       = 20.6%
  Median APE     = 13.5%
  95th pct APE   = 56.4%
  99th pct APE   = 114.5%
  
Plot of predicted vs. actual:
wallclock-by-ml

raswaters (approved):

PR #224, opened 2026-08-18
CALCLOUD-475 - Replace tensorflow with scikit-learn and move Python 3.9 -> 3.11

sashawood:
I apologize for this, but this PR has several changes. It's probably easiest to review them one commit at a time.  If you want, I can pull them apart into smaller PRs.

I have run autotest 08105 with the this set of changes and it works.  Test results are here: https://dsbwww.stsci.edu/dsb_pages/parameterized_auto/562385/robot/regr_08105_ReproInTheCloudEndToEnd_log-20260818-145333.html
The only error is expected - it is FITS file differences related to the calcloud tag being used.

**1. Minor change: [CALCLOUD-475 Remove unnecessary numpy from model_ingest lambda](https://github.com/spacetelescope/calcloud/commit/8ffed90557e94f28717511523a7d06eb301b481f)**
The numpy import has been causing model_ingest lambda to fail for 6 months.  We have not been storing data about actual and predicted resources in DynamoDB during this time.  The fix is to use Python's native round() method in place of numpy.round().

**2. Minor change: [CALCLOUD-475 Check for no job-ids before trying to broadcast](https://github.com/spacetelescope/calcloud/commit/44d5ca4eba8eb3480643c610f73a4e64613f20f0)**
This has been causing harmless errors in our logs, when we try to broadcast information about ongoing jobs, and there are no ongoing jobs.

**3. Major change: [CALCLOUD-475 Move to Python 3.11](https://github.com/spacetelescope/calcloud/commit/631b00602e4747a20fda88977fb1bfef3752331f)**
We were using Python 3.9 (built-in AmazonLinux2023 python) on the ami_rotation EC2 instance and the the ci-node.  We were using Python 3.11 on all the lambdas, including the JobPredict lambda.  Now we're using Python3.11 everywhere

**4. Major change: [CALCLOUD-475 Replace tensorflow with scikit-learn](https://github.com/spacetelescope/calcloud/commit/998a3a59afbdbb8782974ce7dab13e2399aa9274)**

I have replaced the JobPredict code that used tensorflow with new code with an identical interface that uses scikit-learn.

In CALCLOUD-474, I suggested using a RandomForestRegressor for memory and a HistGradientBoostingRegressor for wallclock.  I'd like to change to using the HistGradientBoostingRegressor for both, because the model is much smaller.
RandomForestRegressor has a 45MB model, and HistrgradientBoostingRegressor has a 1MB model.
At 45 MB, I would want to store it in S3, and then we have an annoying versioning issue to handle.
The HistGradientBoostingRegressor is slightly worse than the RandomForestRegressor, but still much better than the existing ML model.

HistGradientBoostingRegressor: 
CALCLOUD-475-memory-HistGradientBoostingRegressor
R²   = 0.962
MAE  = 0.05
RMSE = 0.13
MAPE = 5.5%

RandomForestRegressor:
CALCLOUD-475-memory-RandomForestRegressor
R² = 0.981  
MAE = 0.04  
RMSE = 0.09  
Mean APE = 5.0%

ML model: 
CALCLOUD-475-memory-ML
R² = 0.638  
MAE = 0.25  
RMSE = 0.39  
MAPE = 25.3%

**5. Major change: [CALCLOUD-475 Remove unnecessary training code](https://github.com/spacetelescope/calcloud/commit/2d34b4b95c89a5396b9095be205dcf5254f5efb6)**

Removed:
* the tensorflow model files used for prediction
* the docker image for building the tensorflow model
* the AWS Batch job queue and compute environment for building the tensorflow model
* the code in terraform and deploy files to deploy the docker image for the building tensorflow model

**6. Minor change: [CALCLOUD-475 Get rid of deprecation warnings for terraform](https://github.com/spacetelescope/calcloud/commit/79c36381ff4253994e11a0aa373bb00d3efa89fe)**
Replace "false" with false

**7. Minor change: [CALCLOUD-475 Remove unnecessary docker components](https://github.com/spacetelescope/calcloud/commit/5632601c0dc99f45be9b31e54526792f7be736d4)**
We don't need special handling to STSCI certificates any more.  And we don't need jdk1.8.0

**8. Minor change: [CALCLOUD-475 Switch to using requirements files in Docker build](https://github.com/spacetelescope/calcloud/commit/bf7fbbc90be2723aa190a814c76865f2e71c1ba8)**
Put all requirements in one docker file

**9. Minor change [CALCLOUD-475 Black formatting and flake8 ignore](https://github.com/spacetelescope/calcloud/commit/afc1afc342a99cabbafe842dc51e3976c5357ff3)**

**10. Minor change [CALCLOUD-475 Tag CALCLOUD-475-9](https://github.com/spacetelescope/calcloud/commit/0ab8d3992f50b1b7d41d3398a85fe072c7e3c7d0)**


ami_rotation/ami_rotation_userdata.sh
```diff
@@ -115,7 +116,7 @@ nvm install 16
 npm config set registry http://registry.npmjs.org/
 npm config set cafile /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
 npm install -g awsudo@1.5.0
-python3 -m pip install -q --upgrade pip && python3 -m pip install boto3 -q
+python3.11 -m pip install -q --upgrade pip && python3.11 -m pip install boto3 -q
```
raswaters, line 119: Here and throughout: leave unchanged? Install of python3.11 should leave python and python3 available and pointing to the latest installed version. This will reduce need for future changes.

iac/codebuild/Dockerfile
```diff
@@ -46,7 +36,7 @@ RUN npm install n -g
 RUN npm config set registry http://registry.npmjs.org/ && \
     npm install -g awsudo@1.5.0
 
-RUN pip3 install awscli
+RUN pip3.11 install awscli
```
raswaters, line 39: Same comment as for python3.11, install should link pip3 (and likely pip)

lambda/JobPredict/predict_handler.py
raswaters: This needs an update to remove the reference to neural networks. Grepping around, I found another reference to neural networks in terraform/lambda_job_predict.tf

notes/airflow/airflow_dagger.py
```diff
@@ -8,57 +8,53 @@
 DEFAULT_QUEUE = "hstdp-batch-queue"
 # DEFAULT_JOB_DEFINITION = "hstdp-ipppssoot-job"
 DEFAULT_JOB_DEFINITION = "hstdp-ipppssoot-job-dev"
-DEFAULT_AIRFLOW_EMAIL = "jmiller@stsci.edu"       
+DEFAULT_AIRFLOW_EMAIL = "jmiller@stsci.edu"
```
raswaters: Probably time to update that. Ideally team-violet@stsci.edu, but that may not be accessible from outside. Can we include more than one person? If so, let's list you and me. But is this part even used? This is under the "notes" subdirectory.

notes/airflow/airflow_ipppssoot_dag.py
```diff
@@ -3,34 +3,211 @@
...
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O69H2UBJQ', 'O69H2UBJQ', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/OCO603010', 'OCO603010', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/OD0T5QSYQ', 'OD0T5QSYQ', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/OBC405010', 'OBC405010', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O4DV020A0', 'O4DV020A0', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O60309CHQ', 'O60309CHQ', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O62CBCURQ', 'O62CBCURQ', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O5GJ0ZHYQ', 'O5GJ0ZHYQ', 'stis', 'hstdp-process', 1, 512, 300),
-    ('s3://jmiller-hstdp-output', 'batch-17-2020-01-31T22-18-19/O4XCL7WKQ', 'O4XCL7WKQ', 'stis', 'hstdp-process', 1, 512, 300),
+    (
```
raswaters: In this case, the auto-formatting makes the file trickier to read and especially makes comparison between lines more difficult. This may be a good case to use #yapf: disable or add this file to the yapf ignore file.

terraform/lambda_model_ingest.tf
```diff
@@ -33,7 +33,7 @@ module "lambda_model_ingest" {
     {
       # this is the lambda itself. The code in path will be placed directly into the lambda execution path
       path = "${path.module}/../lambda/ModelIngest"
-      pip_requirements = true
+      pip_requirements = false
```
raswaters: Why was this changed to false? Presumably because there are no requirements anymore? Can this line be deleted?

ami_rotation/ami_rotation_userdata.sh
sashawood: On an EC2 instance using AmazonLinux2023, python3 always goes to the "system Python" which is Python3.9.  
See first 6 paragraphs (they're short) of https://docs.aws.amazon.com/linux/al2023/ug/python.html - "Note: Do not change what the /usr/bin/python3 symlink points to because this might break the core functionality of AL2023."

Calling python3 here is why we're still using Python 3.9 even though we tried to move to Python 3.11.

iac/codebuild/Dockerfile
sashawood: This also needs to remain pip3.11 because this Dockerfile uses AmazonLinux2023: "FROM public.ecr.aws/amazonlinux/amazonlinux:2023"

lambda/JobPredict/predict_handler.py
sashawood: Done

notes/airflow/airflow_dagger.py
sashawood: The notes directory is untouched since 6 years ago.  The only reason it showed up in this PR is because I black formatted everything.  The notes directory has:
1. An Airflow experiment from 6 years ago. 
2. Json representation of the state of AWS resource from 6 years ago for reference.

I think we should actually delete all of this. What do you think?

terraform/lambda_model_ingest.tf
sashawood: It was changed to false because there are no requirements anymore.  This line is not necessary.  I will remove it.

notes/airflow/airflow_ipppssoot_dag.py
sashawood: I you want to keep the notes directory around, I'm happy to do this.

notes/airflow/airflow_dagger.py
raswaters: I have reviewed the contents of the notes directory. This is all old stuff, and the few things that were helpful in the notes have already been absorbed in our code base. If we do want to do more airflow experiments, we would ask the Roman team, not these old notes. If anyone did want to look at these old notes, they'll remain available in git. Please delete the notes directory.

notes/airflow/airflow_ipppssoot_dag.py
raswaters: Nope, I think the notes directory should be deleted.

notes/airflow/airflow_dagger.py
sashawood: I deleted the notes directory.

raswaters (approved):

PR #226, opened 2026-09-08
CALCLOUD-488 Extend kill-time for svm/mvm to 72 hours

sashawood:
Currently, svm's and mvm's are always assigned clockTime = 20 minutes.
The prediction lambda returns clockTime * (1 + std_err) = 20 minutes * (1 + 5) = 120 minutes = 2 hours.
This means the job is killed if it lasts longer than 2 hours.

This is too short, as some mvms on premises take up to 54 hours.

This change sets clockTime = 12 hours.
The prediction lambda returns clockTime * (1 + std_err) = 12 hours * (1 + 5) = 72 hours.

I had to make one additional change.  The _get_resources() function caps the maximum kill time at 48 hours.  I have left this as 48 hours for ipppssoots, but changed to 72 for svm/mvms.

I elected to make this change for both SVMs and MVMs, because we previously did not know about the timeouts when re-processing MVMs, so it is possible that SVMs are hitting the same timeouts.  I plan to collect data on time of SVMs and MVMs in an upcoming build, and then we can re-evaluate.

**Testing**
I ran 08105 - https://dsbwww.stsci.edu/dsb_pages/parameterized_auto/578695/robot/regr_08105_ReproInTheCloudEndToEnd_log-20260908-154219.html. This failed only due to FITS file differences relating to AWSYSVER.

I ran 08106, which succeeded.  https://dsbwww.stsci.edu/dsb_pages/parameterized_auto/578774/robot/regr_08106_HAPInTheCloud_log-20260908-172919.html


calcloud/plan.py
```diff
@@ -111,7 +111,10 @@ def invoke_lambda_predict(dataset, dataset_type, output_bucket):
     else:
         # Return a default 'predictions' for now since models for SVM and MVM not yet implemented
         predictions = dict()
-        predictions["clockTime"] = 20 * 60
+        # For SVM's and MVM's use a predicted clocktime of 12 hours.
+        # When multiplied by 6, this gives a maximum expected clock time of 72 hours.
```
raswaters, line 115: Not sure where the 6 comes from. The limit of 72 appears to come from line 148.

sashawood: This code is confusing - the prediction wallclock time gets manipulated a couple of different times in a couple of different functions.

The 6 comes from line 121, where we multiply the predicted clock time by (1 + wc_std).  wc_std = 5 by default, which you can see in query_ddb.  

For some dataset entries in DynamoDB, there is a wc_std.  If there is no wc_std, we use 5 for it.  I think the wc_std entries are based on timing of multiple runs on-premises (from the blackboard history table), because we never write the wc_std in the calcloud code.

raswaters: Let's try to clarify this a little by adding " (see line 121)" after the 6 on line 115.

sashawood: I have simplified the SVM/MVM logic - there was no need for it to be so complicated.

https://github.com/spacetelescope/calcloud/pull/226/changes#diff-777496f379e7dd8bf236b82ea3b616ad452707e2484da9528b82672fea6c4716R116-R119

calcloud/plan.py
```diff
@@ -108,15 +108,16 @@ def invoke_lambda_predict(dataset, dataset_type, output_bucket):
...
     else:
-        # Return a default 'predictions' for now since models for SVM and MVM not yet implemented
-        predictions = dict()
-        predictions["clockTime"] = 20 * 60
-        predictions["memBin"] = 1
-    # defaults: db_clock=20 minutes, wc_std=5
-    db_clock, wc_std = query_ddb(dataset)
-    clockTime = predictions["clockTime"] * (1 + wc_std)
-    return clockTime, db_clock, predictions["memBin"]
+        # The maximum MVM time that seen on-prem is 54 hours, so set SVM/MVM clockTime to 72 hours.
```
raswaters: Remove "that"

raswaters (approved):

calcloud/plan.py
sashawood: done

PR #227, opened 2026-09-08
CALCLOUD-486 Adjust memory bins to account for overhead

sashawood:
I discovered that the memory bins are not strictly 2, 8, 16, 64 GB.  We reserve 128 MB out of every 1024 MB for use by ECS.  See locals.tf for details.

The lambda function predicts the amount of memory that will be used.  I added 10% to account for errors within the prediction regression.  I needed to lower my bucket sizes by a factor of 128/1024 to account for ECS processing.

This change reduces the bucket sizes in the lambda function so we do not assign jobs to buckets that are too small.

**Testing**
I ran 08105 - https://dsbwww.stsci.edu/dsb_pages/parameterized_auto/578695/robot/regr_08105_ReproInTheCloudEndToEnd_log-20260908-154219.html. This failed only due to FITS file differences relating to AWSYSVER.

I ran 08106, which succeeded.  https://dsbwww.stsci.edu/dsb_pages/parameterized_auto/578774/robot/regr_08106_HAPInTheCloud_log-20260908-172919.html


lambda/JobPredict/predict_handler.py
```diff
@@ -162,12 +162,15 @@ def predict_memory(feature_dict):
 
     predicted_memory = memory_model.predict(feature_df)[0]
 
+    # 128 of 1024 MB is reserved for ECS.  See locals.tf.
```
raswaters, line 165: 128 of each 1024 is for ECS? Looks like that's what locals.tf says. Wondering if that's why we also pad the memory by a factor of 1.1? I.e., was this overhead implicitly accounted for? We could also modify "memory" so we have to do the multiplication only once.

sashawood: When I built the new memory prediction function, I added 10% padding to allow for the error bars that I saw on memory predictions.  If a job was predicted to use 1.9 GB, the actual memory use would be within about 1.7 - 2.1 GB.  I added the 10% padding, so those 1.9GB predictions would less likely to fail.

Then I found in the code the 128 / 1024 padding for ECS, which pre-dates me.  The memory numbers we use as input to our prediction function are for the caldp process only - they do not account for whatever memory the OS requires.  We need to leave extra space for the OS memory.

I thought I should combine both my 10% and the 128/1024.  But that's just a gut feeling. What would you like to do here?

raswaters: I would drop line 166 and instead factor the 1-(128/1024) into "memory" and then leave the lines 169, 171, and 173 as they were originally. Then to the comment on line 165 explain the reason for the 10% padding.

sashawood: Done

raswaters (approved):

PR #228, opened 2026-09-08
CALCLOUD-472 Replace awsudo with AWS_PROFILE

sashawood:
This change removes awsudo everywhere and replaces it with AWS_PROFILE.

awsudo uses a Node.js package that has not been updated since 2024: https://github.com/meltwater/awsudo.
It uses Node.js 16, which was last updated in 2023.   Node.js 16 is flagged by the vulnerability scanner in AWS, and we should remove it.

AWS_PROFILE provides a more robust and standardized solution.

awsudo is used on the ci-node and the ami-rotation node, and nowhere else.  It is not used in any of the AWS lambda or in the batch containers.

I have verified that AWS_PROFILE works correctly all places the the ci-node uses it.  I have also made changes for the ami-rotation node.  I have not verified the ami-rotation changes, because the ami-rotation node is not currently working.  However, when I fix ami-rotation, I will verify that AWS_PROFILE is working as expected there as well.

### DIFF
--- a/ami_rotation/ami_rotation_userdata.sh
+++ b/ami_rotation/ami_rotation_userdata.sh
@@ -109,16 +109,25 @@ yum install python3.11-pip -y -q
 sudo -i -u ec2-user bash << EOF
 mkdir ~/bin ~/tmp
 cd ~/tmp
-curl -s -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
-bash ~/.nvm/nvm.sh
 source ~/.bashrc
-nvm install 16
-npm config set registry http://registry.npmjs.org/
-npm config set cafile /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
-npm install -g awsudo@1.5.0
 python3.11 -m pip install -q --upgrade pip && python3.11 -m pip install boto3 -q
 cd ~
 rm -rf ~/tmp
+EOF
+
+# Set up the AWS Profile for the admin role.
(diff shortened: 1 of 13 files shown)


ami_rotation/ami_rotation_userdata.sh
```diff
@@ -109,18 +109,27 @@ yum install python3.11-pip -y -q
...
+[profile hst_reprocessing_admin_role]
+region = us-east-1
+role_arn = ${admin_arn}
+credential_source = Ec2InstanceMetadata
+
+# This profile is required for ecr describe-image-scan-findings because
+# that command cannot be called cross-account.
+[profile hst-repro-cross-account-ECR-access]
+region = us-east-1
+role_arn = arn:aws:iam::378083651696:role/hst-repro-cross-account-ECR-access
```
raswaters, line 129: Do we want to hardcode this? Can we use something like what we used in terraform/tools/hst_admin_role_shim.sh?

sashawood: hst_admin_role_shim.sh uses a call to STS because it needs to find out what ACCOUNT_ID it should use in the ARN for hst_reprocessing_admin_role.

`HST_ADMIN_ARN="arn:aws:iam::${ACCOUNT_ID}:role/hst_reprocessing_admin_role"`

For this case, the ACCOUNT_ID is always the same - it's the account for the ECR images.  It's the same for all our accounts, and it's appropriate to hard code it.

raswaters (approved):

